PR #2, opened 2022-07-28
Update links to arithmetization.md

sshine:
Some links were broken, others were slightly redundant.

One link was to a specific section of arithmetization.md.

Let's think of other places where such direct links make sense.


sshine:
Merged elsewhere.

PR #29, opened 2022-08-15
Use 'path' as dependency fallback in Cargo.toml

sshine:
Instead of switching between crates.io and a local copy of twenty-first
by commenting in another dependency, specify both `version` and `path`.

This corresponds to having "multiple locations" as described in The
Cargo Book:

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

> It is possible to specify both a registry version and a `git` or
> `path` location. The git or path dependency will be used locally (in
> which case the `version` is checked against the local copy), and when
> published to a registry like crates.io, it will use the registry
> version. Other combinations are not allowed.

Since both the repositories `triton-vm` and `twenty-first` contain a
crate in a sub-directory, one can create a symlink in the right position
to make the relative path `../twenty-first` correspond to the crate from
the local clone on disk:

```
~/Projects/triton-vm/triton-vm $ cd ..
~/Projects/triton-vm $ ln -s ../twenty-first/twenty-first twenty-first
```

When the directory is not present, the crate on crates.io is used, and
when present and correctly positioned, the crate on disk is used. One
can either clone directly from

https://github.com/Neptune-Crypto/twenty-first

or clone from a remote fork of that repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
+# Ignore twenty-first, since this is most likely a symlink to a repository that
+# lives in a different repository. This may eventually become a git submodule
+# or similar.
+twenty-first
+
 ### CLion ###
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
--- a/README.md
+++ b/README.md
@@ -39,3 +39,28 @@ We currently don't recommend using it in production.
 
 Please note that the [Instruction Set Architecture](./specification/isa.md) is not to be considered final.
 However, we don't currently foresee big changes.
+
+## Running the Code
+
+The Rust implementation of Triton VM resides in [triton-vm](./triton-vm) and can be [found on crates.io](https://crates.io/crates/triton-vm).
+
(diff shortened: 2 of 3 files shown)


jan-ferdinand:
That's really cool! Thanks 😊

aszepieniec:
What am I doing wrong?

```
error: failed to get `twenty-first` as a dependency of package `triton-vm v0.1.1 (/home/alan/repos/triton-vm/triton-vm)`

Caused by:
  failed to load source for dependency `twenty-first`

Caused by:
  Unable to update /home/alan/repos/triton-vm/twenty-first

Caused by:
  failed to read `/home/alan/repos/triton-vm/twenty-first/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

sshine:
> ```
> Caused by:
>   failed to read `/home/alan/repos/triton-vm/twenty-first/Cargo.toml`
> ```

I don't know, but one guess is that the git root was symlinked rather than the crate within.

```
$ cd /home/alan/repos/triton-vm
$ ln -s ../twenty-first/twenty-first twenty-first
```

rather than `ln -s ../twenty-first twenty-first`.

sshine:
@aszepieniec: My apologies, I had misunderstood that once `path` is present, it will always prefer `path` locally.

I changed it back to start out with `path` commented out.

Let me know if you have other problems.

PR #35, opened 2022-08-17
Add Triton Improvement Proposals 0001 - 0004

aszepieniec:
TIP 0001 is a writeup that introduces a *contiguity argument*, which is a Polynomial IOP / RAP hybrid for establishing that the set of values is distributed contiguously across an evaluation domain. Phrased crudely, all *a*'s are next to each other, for any *a*. This construction addresses one issue in a proposed fix for the memory-consistency problem.

(_slight edit by @jan-ferdinand_)


tips/tip-0001/tip-0001.md
```diff
@@ -0,0 +1,67 @@
...
+
+The running product starts with $g$ initially and accumulates a factor $X - \mathsf{ramp}$ in every pair of rows where $\mathsf{ramp} \neq \mathsf{ramp}^\star$. This evolution corresponds to one transition constraint: $(\mathsf{ramp}^\star - \mathsf{ramp}) \cdot (\mathsf{rp}^\star - \mathsf{rp} \cdot (X - \mathsf{ramp})) + (1 -(\mathsf{ramp}^\star -\mathsf{ramp}) \cdot \mathsf{di}) \cdot (\mathsf{rp}^\star - \mathsf{rp})$.
+
+The associated terminal value is $T_{\mathsf{rp}}$ and the terminal constraint is stipulates that the factor $X - \mathsf{ramp}$ from the last row is accumulated as well: $\mathsf{rp} \cdot (X - \mathsf{ramp}) - T_{\mathsf{rp}}$. Alternatively, if the last row is a padding row and should not be accumulated into the running product, then the last value in the column is already the terminal value: $\mathsf{rp} - T_{\mathsf{rp}}$.
```
jan-ferdinand, line 37: In Triton, the memory tables can easily be derived after padding has happened, implying that any permutation arguments between the processor and the respective memory table will work out. Are you presenting padding rows here in order to make the contiguity argument more generic, i.e., easier to apply beyond Triton VM?

tips/tip-0001/tip-0001.md
```diff
@@ -0,0 +1,67 @@
...
+
+$$ \mathsf{fd} \mapsto \mathsf{fd}^ \star = (X - \mathsf{ramp}) \cdot \mathsf{fd} + \mathsf{rp}   . $$
+
+This update rule is called the *product rule of differentiation* because when ignoring $g(X)$ and $h(X)$ or when setting $h(X) = g'(X)$ and assuming $\mathsf{ramp}^\star \neq \mathsf{ramp}$, then
+
+$$ \frac{\mathsf{d}  \mathsf{rp}^\star}{\mathsf{d}   X} = \frac{\mathsf{d}  (X - \mathsf{ramp}) \cdot \mathsf{rp}}{\mathsf{d}   X} $$
+$$ = (X - \mathsf{ramp}) \cdot \frac{\mathsf{d}   \mathsf{rp}}{\mathsf{d}   X} + \frac{\mathsf{d}  ( X - \mathsf{ramp})}{\mathsf{d}   X} \cdot \mathsf{rp} $$
+$$ = (X - \mathsf{ramp}) \cdot \mathsf{fd} +\mathsf{rp} \enspace . $$
+
+The transition constraint for $\mathsf{fd}$ is $(\mathsf{ramp}^\star - \mathsf{ramp}) \cdot (\mathsf{fd}^\star - \mathsf{rp} - (X - \mathsf{ramp}) \cdot \mathsf{fd}) + (1 -(\mathsf{ramp}^\star -\mathsf{ramp}) \cdot \mathsf{di}) \cdot (\mathsf{fd}^\star - \mathsf{fd})$.
```
jan-ferdinand, line 49: This is a personal preference and I'm open to counter arguments: I'm adding $\mathsf{di}$ as a factor to the first summand here to
1. make the two summands symmetric, and
1. keep the leading term's constant unchanged.

The degree of the constraint polynomial does not increase, but implementing the `extend` function becomes (arguably marginally) simpler.

tips/tip-0001/tip-0001.md
```diff
@@ -0,0 +1,67 @@
...
+
+**Zero-Knowledge.** The prover sends four polynomial evaluations, two of which are terminals. We treat each value in turn.
+ 1. $T_{\mathsf{rp}}(\alpha)$: For any $T_{\mathsf{rp}}(\alpha)$, for any list $\lbrace \mathsf{ramp}_i \rbrace_i$ of accumulated $\mathsf{ramp}$ values, and for any challenge $\alpha$, there is an initial $g$ that is consistent with the given view. Indeed, the mapping from $g \mapsto T_{\mathsf{rp}}(\alpha)$ is affine and the coefficient is nonzero as long as $\alpha \not \in \lbrace\mathsf{ramp}_i\rbrace_i$.
+ 2. $T_ {\mathsf{fd}}(\alpha)$: For any pair $(T_ {\mathsf{rp}}(\alpha), T_ {\mathsf{fd}}(\alpha))$, for any list $\lbrace \mathsf{ramp}_ i \rbrace_ i$ of accumulated $\mathsf{ramp}$ values, and for any challenge $\alpha$, there is an initial $g$ that is consistent with the given view. Indeed, the mapping from $h \mapsto T_ {\mathsf{fd}}(\alpha)$ is affine and the coefficient is nonzero as long as $\alpha \not \in \lbrace\mathsf{ramp}_ i\rbrace_ i$.
```
jan-ferdinand, line 65: Is the initial in question really $g$, not $h$?

tips/tip-0001/tip-0001.md
```diff
@@ -0,0 +1,67 @@
...
+
+Until the verifier supplies the challenge $\alpha$, the terminal "values" are in fact polynomials: $T_{\mathsf{rp}}(X)$ and $T_{\mathsf{fd}}(X)$. The polynomials $a(X)$ and $b(X)$ are *randomized* Bezout coefficients of the relation
+$$ a(X) \cdot T_{\mathsf{rp}}(X) + b(X) \cdot T_{\mathsf{fd}}(X) = \gcd(T_{\mathsf{rp}}(X),T_{\mathsf{fd}}(X)) \enspace .$$
+If zero-knowledge is an objective, then $a(X) = a^*(X) + k \cdot T_{\mathsf{fd}}(X)$ and $b(X) = b^*(X) - k \cdot T_{\mathsf{rp}}(X)$, where $a^*(X)$ and $b^*(X)$ are the minimal-degree Bezout coefficients as returned by the extended Euclidean algorithm, and where $k \in \mathbb{F}$ is a uniformly random field element called the *Bezout randomizer*. For the sake of the soundness analysis, set the degree bound on $a(X)$ and $b(X)$ to $T$, the height of the table.
```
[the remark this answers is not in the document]
jan-ferdinand: Consequently, we could call the initials $\mathsf{example}_0$, clearly indicating which initial belongs to which column.

jan-ferdinand (changes requested):

yczhangsjtu (commented):
What is the prover complexity? In particular, to apply the extended Euclidean algorithm, does the prover need to explicitly compute and store $T_{fd}(X)$ and $T_{rp}(X)$? And naively applying the extended Euclidean requires repeated polynomial division. What would be the cost of that?

tips/tip-0001/tip-0001.md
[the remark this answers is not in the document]
aszepieniec: I like this notation and will use it from now on. I wish I had thought of it for the Brainfuck tutorial but that I won't update.

tips/tip-0001/tip-0001.md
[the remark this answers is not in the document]
aszepieniec: For genericity, yes.

tips/tip-0001/tip-0001.md
aszepieniec: The one in bullet point 2 should be $h$.

aszepieniec:
> What is the prover complexity? In particular, to apply the extended Euclidean algorithm, does the prover need to explicitly compute and store T_{fd}(X) and T_{rp}(X)? And naively applying the extended Euclidean requires repeated polynomial division. What would be the cost of that?

I was surprised that the complexity of xgcd is difficult to determine. I found [this source](https://math.stackexchange.com/questions/258596/what-is-the-time-complexity-of-euclids-algorithm-upper-bound-lower-bound-and-a), which claims that the complexity for integer gcd is $O(\log a)$ where $a$ is the smaller integer. I assume that polynomial xgcd has a complexity of $O(d)$ or $O(d \log d)$ at worst.

The polynomial needs to represent the polynomials $T_ {fd}(X)$ and $T_ {rp}(X)$ explicitly in order to do the xgcd. To get these polynomials, the most straight forward way is to multiply all factors $(X - \mathsf{ramp}_ i)$ in frequency domain, followed by an inverse NTT. The cost is $O(F \cdot N \log N)$ field operations where $F$ is the number of factors and $N$ is the size of the NTT domain. Note that in practical execution traces, the number of factors $F$ is much smaller than the length of the execution trace.

tips/tip-0001/tip-0001.md
aszepieniec: I think the degree of the constraint polynomials do increase. That was exactly my motivation for dropping the redundant factor.

aszepieniec:
I found [this paper](https://link.springer.com/content/pdf/10.1007/978-3-662-53887-6_7.pdf) which says §4.2:

> It is well-known that the complexity for finding the GCD of two polynomials of degree $d$ is $O(d \log_2 d)$.

aszepieniec:
About computing $\mathsf{rp}_T(X)$ and $\mathsf{fd}_T(X)$ we need to accumulate $O(T)$-many linear factors. That sounds quadratic, no?

No! The polynomial $\prod_i (X - z_i)$ is in fact a zerofier for $\lbrace z_i \rbrace$. Zerofiers can be computed in $O(N \log^2 N)$ time using `fast_zerofier` from the [anatomy](https://aszepieniec.github.io/stark-anatomy/faster).

yczhangsjtu:
It seems that `fast_zerofier` requires the domain to be a multiplicative subgroup. Are $\{z_i\}_i$?

tips/tip-0001/tip-0001.md
jan-ferdinand: You are right, it does increase the degree. My sense for symmetry is bummed out, but I withdraw the suggestion.

jan-ferdinand:
I think I'm in favor of having the TIPs available on the master branch. Navigating branches as an additional dimension to navigating the filesystem is not necessarily obvious to individuals new to the project. Squashing & Rebasing should help with not polluting the commit history of the master branch.

jan-ferdinand (approved):

jan-ferdinand:
Superseded by #79.

PR #66, opened 2022-09-20
Pad all tables to the same height

### LINKED ISSUE #10: Pad all tables to the same height
A Verifier needs to compute (what we call) `unit_distances` from the extension tables' `padded_height`s. For that, they currently need to have access to the padded height of all tables, which (for now) is communicated via the proof stream.

At a runtime cost for general proving, moving to a single `padded_height` simplifies the Verifier. A simpler Verifier implies a faster runtime for the recursive Verifier. Thus, the proposed change should be implemented.

(Originally issue number 31 in the internal issue tracker.)

jan-ferdinand:
Fix #10

### DIFF
--- a/triton-vm/src/table/jump_stack_table.rs
+++ b/triton-vm/src/table/jump_stack_table.rs
@@ -315,19 +315,6 @@ impl JumpStackTable {
 }
 
 impl ExtJumpStackTable {
-    pub fn with_padded_height() -> Self {
-        let matrix: Vec<Vec<XFieldElement>> = vec![];
-
-        let inherited_table = Table::new(
-            BASE_WIDTH,
-            FULL_WIDTH,
-            matrix,
-            "ExtJumpStackTable".to_string(),
-        );
-
-        Self { inherited_table }
-    }
-
     pub fn ext_codeword_table(
         &self,
         fri_domain: &FriDomain<XFieldElement>,
--- a/triton-vm/src/table/jump_stack_table.rs
+++ b/triton-vm/src/table/jump_stack_table.rs
@@ -1,7 +1,7 @@
 use itertools::Itertools;
 use num_traits::{One, Zero};
 use twenty_first::shared_math::b_field_element::BFieldElement;
-use twenty_first::shared_math::mpolynomial::MPolynomial;
+use twenty_first::shared_math::mpolynomial::{Degree, MPolynomial};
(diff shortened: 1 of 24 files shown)


triton-vm/src/table/table_collection.rs
```diff
@@ -67,49 +75,135 @@ fn to_vec_vecs<T: Sized + Clone, const S: usize>(vector_of_arrays: &[[T; S]]) ->
...
+    debug_assert!(
+        0 == padded_height || is_power_of_two(padded_height),
+        "The padded height was: {}",
+        padded_height
+    );
+    DataPF::primitive_root_of_unity(padded_height).unwrap()
+}
+
+/// Returns the relation between the FRI domain and the omicron domain
+pub fn unit_distance(padded_height: usize, omega_order: usize) -> usize {
```
aszepieniec, line 92: Since `padded_height` is never zero, I don't think this function adds anything. We could simply inline the division `omega_order / padded_height` in every place where this function is invoked, ad put the result of that division into a variable named so as to describe its purpose.

jan-ferdinand: Fixed in cd1c8e9572c0c855821563b7d0a0b6086f9b1813

triton-vm/src/table/jump_stack_table.rs
```diff
@@ -305,16 +315,12 @@ impl JumpStackTable {
 }
 
 impl ExtJumpStackTable {
-    pub fn with_padded_height(num_trace_randomizers: usize, padded_height: usize) -> Self {
+    pub fn with_padded_height() -> Self {
```
aszepieniec, line 318: Is this function (and similar functions for other tables) used anywhere? It seems strange to have a function name that anticipates an argument and then not supply that argument. Also, I noticed that the same function is defined on `ExtTableCollection` but with the argument there.

jan-ferdinand: Good catch. This might be a remnant of some previous refactoring effort. To be frank, I lost track of some of the bits during the refactor for this PR.

jan-ferdinand: Fixed by 59e5b32

sshine (approved):
Looks good. I only had comments about formatting code, so I discarded those for now.

jan-ferdinand:
Rebased, squashed, and merged.

PR #73, opened 2022-09-27
Derive Bounded and From<*TableColumn> for usize using EnumIter (strum)

### LINKED ISSUE #17: Derive Ext*TableColumn indexing
(Originally posted by @sshine.)

Right now, each `*TableColumn`'s `.into()` for `usize` hardcodes enumeration, and the corresponding `Ext*TableColumn`'s `.into()` for `usize` does, too, starting with a constant that can go out of sync.

Instead, have two `*Column` that are simple enums (not sum types), derive their `.into()`s, and combine them in a third sum type that piggy-backs on the derived instances. This also makes property tests ("for all columns") possible.

The main change here is that deriving enum instances doesn't work for generic sum types, only for plain enum types (sum types without constructor arguments).

(Originally issue number 38 in the internal issue tracker.)

sshine:
This addresses and closes #17.

Changing the columns (adding, removing, re-ordering) does not cause sporadic indexing errors because some trait instance for some column wasn't updated.

For each table, enums exist that enable addressing a table row by the name of the column rather than its arbitrary number in the column order. This makes the VM and its constraint polynomials more readable and less brittle.

By changing the enums so there are two simple ones, e.g.:
- `BaseProcessorTableColumn` (containing base columns)
- `ExtProcessorTableColumn` (containing ext columns)

and one combined one:

```rust
pub enum ProcessorTableColumn {
    Base(BaseProcessorTableColumn),
    Ext(ExtProcessorTableColumn),
}
```

When `BaseProcessorTableColumn` and `ExtProcessorTableColumn` derive `EnumIter`, it means that `From` and `Bounded` can be implemented in terms of a derived order, rather than one that is manually specified.

This was not possible when `ExtProcessorTableColumn` had a constructor that took an argument, like so:

```rust
pub enum ExtProcessorTableColumn {
    BaseColumn(ProcessorTableColumn),
    ...
}
```

because the `EnumIter` derivation only works for simple enums without arguments. The combined `ProcessorTableColumn`'s trait instances now relies on the derived instances of its inner parameters.

### DIFF
--- a/triton-vm/src/table/table_column.rs
+++ b/triton-vm/src/table/table_column.rs
@@ -5,11 +5,11 @@
 // --------------------------------------------------------------------
 
 use num_traits::Bounded;
-use std::fmt::{Display, Formatter};
-use HashTableColumn::*;
+use strum::{EnumCount, IntoEnumIterator};
+use strum_macros::{Display, EnumCount as EnumCountMacro, EnumIter};
 
-#[derive(Debug, Clone, Copy)]
-pub enum ProcessorTableColumn {
+#[derive(Display, Debug, Clone, Copy, PartialEq, Eq, EnumIter, EnumCountMacro)]
+pub enum ProcessorBaseTableColumn {
     CLK,
     IsPadding,
     IP,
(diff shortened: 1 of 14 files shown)


triton-vm/src/table/table_column.rs
```diff
@@ -201,39 +187,37 @@ impl std::fmt::Display for ExtProcessorTableColumn {
...
 }
 
-impl From<ExtProcessorTableColumn> for usize {
-    fn from(c: ExtProcessorTableColumn) -> Self {
-        use ExtProcessorTableColumn::*;
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProcessorTableColumn {
+    Base(BaseProcessorTableColumn),
+    Ext(ExtProcessorTableColumn),
+}
```
sshine, line 194: This combined enum seems entirely unnecessary.

We only appear to be using the `*TableColumn` types for
- conversion to `usize` when indexing rows
- conversion via `Display` in the stepping debugger

There appears to be no context where we interchangeably refer to either base columns and extension columns. That could have been the case for constraint polynomials that use the extension columns and the base columns, but in that case, the `usize::from(...)` is sufficient on each enum separately.

jan-ferdinand:
Nice!
Mind pushing this to a feature branch on this repo so I can continue?

jan-ferdinand (approved):

sshine (commented):
Before merging this,
- [x] Get rid of the combined `enum ProcessorTableColumn`
- [x] Derive existing `*TableColumn` enums the same way as processor table.
- [ ] Address all occurrences of `X as usize` vs. `usize::from(X)`.

sshine:
As @jan-ferdinand suggests, we wait until #68 is merged before applying the change to all tables.

sshine:
This PR can be squashed now.

PR #75, opened 2022-10-01
Add some `par_iter`s for a 16 % speedup

Sword-Smith:
With this change the time spent to calculate transition quotients for the processor table drops from 119s to 103s on my DELL XPS with 16 GB RAM.


jan-ferdinand:
Interestingly, I'm observing a slowdown, not a speedup.

| | run 1 | run 2 |
|:---|---:|---:|
| just `iter()` | 94.12s | 92.03s |
| `par_iter()` | 98.40s | 98.57s |

Can someone else provide more data points?

Sword-Smith:
If this speedup is not consistent across various systems, I would not merge it.

jan-ferdinand:
After getting a third opinion on the matter, it seems this is not consistently speeding up the computation across machines. Furthermore, there are significant improvements of the concerned methods coming.

PR #76, opened 2022-10-06
add Alexander's idea for using pure AET + AIR Bézout coefficients

jan-ferdinand:
@AlexanderLemmens has come up with a beautiful, pure AET + AIR solution to achieve contiguous memory regions. This PR incorporates the suggested changes into TIP 0001.


aszepieniec:
Well done! I think it works. Moreover, it is simpler, and it avoids the awkward translation for zero-knowledge. I see only upsides and recommend immediate merger and implementation.

PR #81, opened 2022-10-09
Update triton-vm to use latest twenty-first

sshine:
Update triton-vm to use latest twenty-first
    
- Neptune-Crypto/twenty-first#36: ::variables() had its second argument removed.
- Neptune-Crypto/twenty-first#42: GetRandomElements has been removed.

### DIFF
--- a/triton-vm/Cargo.toml
+++ b/triton-vm/Cargo.toml
@@ -26,7 +26,7 @@ version = "1"
 default-features = false
 
 [dependencies]
-twenty-first = { version = "0.2.0" } # , path = "../twenty-first" }
+twenty-first = { git = "https://github.com/Neptune-Crypto/twenty-first.git", branch = "master" }
 anyhow = "1.0"
 bincode = "1.3"
 blake3 = "1.2"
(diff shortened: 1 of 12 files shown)


triton-vm/Cargo.toml
```diff
@@ -26,7 +26,7 @@ version = "1"
 default-features = false
 
 [dependencies]
-twenty-first = { version = "0.2.0" } # , path = "../twenty-first" }
+twenty-first = { git = "https://github.com/Neptune-Crypto/twenty-first.git", branch = "random-elements" }
```
sshine, line 29: Before merging the present PR, merge this branch on twenty-first and change the dependency to the master branch.

jan-ferdinand: Why not make a new release on crates.io and depend on that? My impression was that depending on git repositories was unusual.

jan-ferdinand:
An additional feature present in the latest version of twenty-first is, afaik, `inverse_or_zero()` of field elements. It might be nice to start calling those methods where it makes sense. However, I'm not sure this PR is the right place for that change.

triton-vm/Cargo.toml
sshine: I'd like to do that once the digest type has been added to twenty-first.

sshine:
> It might be nice to start calling those methods where it makes sense.

That would be a nice addition. This PR isn't doing much, I've added it as a separate commit.

PR #84, opened 2022-10-12
replace `skiz` by `if_then_call`

jan-ferdinand:
As discussed in [this issue](https://github.com/anoma/zkp-compiler-shootout/issues/16), implementing a proper “if then else” is not currently very straightforward using Triton assembly. The new instruction `if_then_call` alleviates the problem.

For instruction `if_then_call`, note the difference between the potential offsets added to instruction pointer `ip` and jump origin `o`. Specifically, if the top of the stack `a` is 0, the instruction pointer `ip` is incremented by 2, skipping the instruction's argument `d`. However, if the top of the stack `a` is not 0, the return address put on the jump stack is `ip+4`. This skips not only the instruction's argument `d`, but also the next two instructions when `return`ing. Together, this mechanic allows following instruction `if_then_call` + `d` with instruction `call` + `e` to serve as the “else” branch. If no “else” branch is desired, instruction `if_then_else` + `d` can be followed with two instructions `nop`. Of course, instruction `if_then_call` can be followed by any instruction. The appropriate care should be taken to `return` only to where it makes sense.

We might drop instruction `skiz` from the ISA, but the combination of `skiz recurse return` is beautiful.


jan-ferdinand:
Expressing `skiz` using `if_then_call`:

```
push 0 eq
if_then_call dummy-label
instruction_to_maybe_skip
nop_or_argument

dummy-label: return
```

Note that the emulated `skiz`, unlike the currently implemented version, does not pop the top of the stack. The technical reason is that any one instruction may change at most one memory pointer. Since `if_then_call` might change the jump stack pointer `jsp`, the length of the operational stack cannot change at the same time, because that would modify `osp`.

aszepieniec:
I thought about emulating if-then-else with `skiz`, and came up with a way to do that. If I understand correctly, the limitation was that the two branches could not leave the stack in an arbitrary state. This construction alleviates that limitation, with no change to the instruction set or arithmetization.

```
dup0
skiz
call if-wrapper
call else-wrapper
pop

if-wrapper:
pop
call if-branch
push 1
return

else-wrapper:
push 0 eq # logical not
skiz
call else-branch
return
```

To verify the execution, see the following table:

| instruction | stack | stack |
|-|-|-|
| `dup0` | `_ 0` | `_ 1` |
| `skiz` | `_ 0 0` | `_ 1 1` |
| `call if-wrapper` | | `_ 1` |
| `pop` | | `_` |
| `call if-branch` | | `_` |
| `... return` | | `&` |
| `push 1` | | `&` |
| `return` | | `& 1` |
| `call else-wrapper` | `_ 0` | `& 1` |
| `push 0 eq` | `_ 0 ` | `& 1` |
| `skiz` | `_ 1` | `& 0` |
| `call else-branch` | `_` | |
| `... return` | `%` | |
| `return` | `%` | `&` |

jan-ferdinand:
Very elegant. This warrants a more thorough analysis of the costs of the two instructions `skiz` and `if_then_call`.

sshine:
A variation of this if-then-else wrapper that I derived is like this:

```
// Before: _ c
// After: _ ...
some-routine:
  push 1 // -> _ c 1
  swap1  // -> _ 1 c
  skiz   // -> _ 1
    call then-branch
  skiz   // -> _
    call else-branch
  ...
  return

// Before: _ 1
// After: _ ... 0
then-branch:
  pop
  ...
  push 0
  return

// Before: _
// After: _ ...
else-branch:
  ...
  return
```

I.e., start by pushing a 1 that will make the else-skiz trigger, and in the then-branch change that 1 to a 0.

A limitation of `call`ing to a conditional branch is that the branch cannot `return` all the way back.

aszepieniec:
> A limitation of calling to a conditional branch is that the branch cannot return all the way back.

I don't follow. Please elaborate.

sshine:
> > A limitation of calling to a conditional branch is that the branch cannot return all the way back.
>
> I don't follow. Please elaborate.

If you want to make an early return in a branch with more than just that return, like

```
// pseudo-syntax
fn foo(c) {
    if c {
        ...x;
        return // leave 'foo'
    }
    ...y
}
```

then the following translation to tasm will not work:

```
foo:
    skiz call foo-then
    ...y
    return // leave 'foo'

foo-then:
    ...x
    return // return to 'foo', does not leave 'foo'
```

since `foo-then`'s `return` only returns to `foo`, whereas the pseudo-syntax leaves `foo`.

A solution to allowing for leaving a function entirely inside a `skiz call` branch is to leave a boolean on the stack (similar to the if-then-else construction) and `skiz return` immediately after returning from the branch.

jan-ferdinand:
There seems to be no immediate need to replace instruction `skiz`.

PR #92, opened 2022-10-20
Flip sign of UniqueClockJumpDiffDiffInverse in proc base table

Sword-Smith:
Subtle sign error here.


aszepieniec:
Great job tracking this :lady_beetle: down :clap: :clapper: 

I hesitate to merge it immediately because there is still a discrepancy between the tests `triton_prove_verify_test` and `triton_table_constraints_evaluate_to_zero_test` that causes the tables to be different. Had that discrepancy not been there, I would have caught the bug you identified while writing the constraints. As long as the discrepancy is there, future constraint-writers fall prey to the same pitfall. Lastly, it might indicate a potentially exploitable soundness error.

aszepieniec:
Cherry-picked in eb9759399e4d84903ec8e3dab2c069c6c98e483d, following the fix of this bug's counterpart and the introduction of tests that would have caught it.

PR #93, opened 2022-10-21
Increase triton-vm to 0.3.1

sshine:
Increase triton-vm to version 0.3.1

Bigger changes:
- Use AlgebraicHasher instead of simple_hasher::Hasher.
- Remove `H: Hasher` type parameter from ProofItem: It was only needed for FriProof, which contains digests, but `Digest` is now a concrete type and not an associated type of Hasher (`H::Digest`).
- Adjust Fri/FriDomain after Neptune-Crypto/twenty-first#41 optimization:

  Let offset and omega be BFieldElements in all cases, and add additional constraints:

  ```
  FF: FiniteField + From<BFieldElement> + MulAssign<BFieldElement>
  ```

  This ensures that the optimized NTT's interface is compatible with FriDomain at a relatively small loss to abstraction.

Smaller changes:
- Rename all FiniteField type parameters to FF
- Fix all linting errors (including tests)
- Remove `impl Default for ProofItem`: This was and should not be necessary.

### DIFF
--- a/triton-vm/Cargo.toml
+++ b/triton-vm/Cargo.toml
@@ -26,7 +26,7 @@ version = "1"
 default-features = false
 
 [dependencies]
-twenty-first = { git = "https://github.com/Neptune-Crypto/twenty-first.git", rev = "0f5374fda" }
+twenty-first = { git = "https://github.com/Neptune-Crypto/twenty-first.git", branch = "master" }
 anyhow = "1.0"
 bincode = "1.3"
 blake3 = "1.2"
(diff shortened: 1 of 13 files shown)


Sword-Smith:
Looks very good to me. Lots of nice cleanup code here.

Sword-Smith:
Now that we have the option of multiplying an X field element with a B field element, there might be possibilities for speedups multiple places. Maybe we should also have the option of evaluating an X field element in a `MPlynomial<BFieldElement>`, since this amounts to the same thing. There might be a factor 4 speedup for many of the transition constraints waiting there.

PR #103, opened 2022-11-03
Serializable Proof Stream

aszepieniec:
This PR separates the "proof" from the "proof stream" notions. In particular:

 - introduce a new standalone `Proof` type, which is a vector `BFieldElements` under the hood but with extra functionality for parsing and storing
 - introduce a new `ProofStream` class that stores vectors of `BFieldElements` rather than `ProofItem`s, but without changing the interface
 - introduce a new trait `BFieldCodec` which captures the functionality that some data structures must be encoded as strings of `BFieldElements` or decoded therefrom

It is a **work in progress**:

 - [x] define trait `BFieldCodec`
 - [x] implement `BFieldCodec` for `ProofItem` and related data structures
 - [x] override `BFieldCodec` for `Vec<BFieldElement>` for better performance (optional)
 - [x] write new `ProofStream`
 - [x] integrate new `ProofStream` into prover
 - [x] tests
   - [x] positive
   - [x] randomized
   - [x] negative

Please note that deserialization is an attack vector. If the attacker exploits a deserialization bug, he may be able to sidestep the cryptgraphic verification machinery altogether. It is therefore imperative that deserialization is bug-free.


aszepieniec:
The file `bfield_codec.rs` is a candidate for migration to twenty-first.

aszepieniec:
These edits now also live on #97. So I'm closing this PR to avoid having to rebase again.

PR #118, opened 2022-11-14
fix a few typos in specs

contrun:


specification/src/copy-constraints.md
```diff
@@ -21,11 +21,11 @@ To establish that the product is correctly computed, the Prover needs to supply
 
 **Protocol Permutation**
  - Verifier is already in possession of polynomial oracles $[t(X)]$ and $[s(X)]$ that agree with $T$ and $S$ on a subgroup $H$ of order $2^k$.
- - Verifier samples $\beta \xleftarrow{\$} \mathbb{F}$ and sends it to Prover.
+ - Verifier samples \(\beta \xleftarrow{\$} \mathbb{F}\) and sends it to Prover.
  - Prover computes running product polynomials, $t_ p(X)$ and $s_ p(X)$ of degree at most $|H|-1$ and sends them to Verifier.
  - Prover and Verifier run an AIR sub-protocol to establish that the AIR is satisfied for $(t(X), s(X), t_ p(X), s_ p(X))$.
 
-What if $T$ and $S$ have multiple columns, say $r$-many? In this case the verifier simply supplies $r$ weights $\alpha_0, \ldots, \alpha_{r-1} \xleftarrow{\$} \mathbb{F}$. The protocol is then applied to the weighted sum of columns.
+What if $T$ and $S$ have multiple columns, say $r$-many? In this case the verifier simply supplies $r$ weights \(\alpha_0, \ldots, \alpha_{r-1} \xleftarrow{\$} \mathbb{F}\). The protocol is then applied to the weighted sum of columns.
```
sshine, line 28: These are not typos. The spec consistently uses `$` as the inline MathJax symbol.

sshine (commented):
Thank you for noticing the wrong hash length here.

specification/src/copy-constraints.md
contrun: Yeah. But it is not rendering correctly in https://triton-vm.org/spec/copy-constraints.html. I wonder what is wrong.

sshine: I'll have a look. Thank you for noticing.

sshine (approved):
@contrun: It turns out to be a KaTeX bug: While `\$` is the right way to escape a dollar sign, somehow KaTeX captures this and produces a literal `$` which then messes up the inline comment. The `\(`s don't seem to fix it, but adding another backslash does. In LaTeX this would mean "linebreak followed by beginning of inline equation", which would be totally wrong.

sshine:
@jan-ferdinand, @aszepieniec: Beware that `$` is a dodgy character when writing KaTeX.

PR #126, opened 2022-11-16
Fix PartialAuthenticationPath Decoding Error

aszepieniec:
There was a bug in the decoding procedure for `Vec<PartialAuthenticationPath>` . This PR fixes the bug.

The culprit was this line in `BFieldCodec`:
```
if (vec_len != 0 || mask != 0) && index == str.len() {
```

This if-statement generates an error if the predicate evaluates to true. The idea is that if the vector length (`vec_len`, number of `Option<Digest>`s in this partial authentication path) is nonzero, or if the mask (`mask`, whose bits indicate whether the option is `Some(..)` or `None`) is nonzero, then we should be expecting at least one digest to follow; and so in this case the index cannot point to the end of the string.

The logic is flawed because the vector length can nonzero with no following digests. This occurs in partial authentication paths consisting entirely of `None`s. The `mask` variable will be zero, thus triggering the predicate erroneously.

The fix is to change the offending line to this:
```
if vec_len != 0 && mask != 0 && index == str.len() {
```

In other words, we are only expecting to read another digest if a) the vector has nonzero length, _and_ b) the mask is nonzero.

### DIFF
--- a/triton-vm/src/stark.rs
+++ b/triton-vm/src/stark.rs
@@ -1749,23 +1749,68 @@ pub(crate) mod triton_stark_tests {
     #[test]
     fn triton_prove_verify_halt_test() {
         let code_with_input = test_halt();
-        let mut profiler = Some(TritonProfiler::new("prove halt"));
         let (stark, proof) = parse_simulate_prove(
             &code_with_input.source_code,
             code_with_input.input.clone(),
             code_with_input.secret_input.clone(),
-            &mut profiler,
+            &mut None,
         );
-        let mut profiler = profiler.unwrap();
-        profiler.finish();
-        let report = profiler.report();
-        println!("{report}");
-
-        println!("between prove and verify");
 
         let result = stark.verify(proof, &mut None);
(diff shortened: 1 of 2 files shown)


triton-vm/src/stark.rs
```diff
@@ -1770,6 +1763,58 @@ pub(crate) mod triton_stark_tests {
         assert!(result.unwrap());
     }
 
+    #[test]
+    #[ignore = "used for catching deserialization errors"]
```
jan-ferdinand, line 1767: Why ignore the tests?

jan-ferdinand (approved):
That looks like a nasty one. Thanks for fixing it!

triton-vm/src/stark.rs
aszepieniec: These tests do not catch errors that `triton_prove_verify_halt` does not already catch. They are useful only for tracking them down. I guess I should replace the comment "catching" with "tracking".

PR #138, opened 2022-11-22
Add option of printing profile output as-you-go

Sword-Smith:
By setting the environment variable `PROFILE_AS_YOU_GO` you can now get the names from the profiler printed as the program runs. This will not be indented nicely like the final report, but this commit does not touch how the final report is printed, so you still get that.

This is introduced because it's nice to have for very long programs, some of our benchmarks/tests still take hours on the fastest computers and it's nice to be able to see during the run how far it has come.

You can test the functionality by running this command:
`PROFILE_AS_YOU_GO=1 cargo t triton_prove_verify_halt_test -- --nocapture`

### DIFF
--- a/triton-profiler/src/triton_profiler.rs
+++ b/triton-profiler/src/triton_profiler.rs
@@ -174,8 +174,8 @@ impl TritonProfiler {
             task_type,
         });
 
-        if std::env::var("PROFILE_AS_YOU_GO").is_ok() {
-            println!("start: {name}");
+        if std::env::var(GET_PROFILE_OUTPUT_AS_YOU_GO_ENV_VAR_NAME).is_ok() {
+            println!("stop: {name}; took {now:.2?}");
         }
     }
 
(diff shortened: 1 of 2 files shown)


triton-profiler/src/triton_profiler.rs
```diff
@@ -219,10 +225,14 @@ impl TritonProfiler {
...
 
     fn plain_stop(&mut self) {
-        let index = self.stack.pop().unwrap().0;
+        let (index, name) = self.stack.pop().unwrap();
         let now = self.timer.elapsed();
         let duration = now - self.profile[index].time;
         self.profile[index].time = duration;
+
+        if std::env::var(GET_PROFILE_OUTPUT_AS_YOU_GO_ENV_VAR_NAME).is_ok() {
+            println!("stop: {name}");
```
jan-ferdinand, line 234: How about also adding the time it took to run the thing, even without pretty formatting in place?
```suggestion
            println!("stop: {name} (took {duration:.2?})");
```

jan-ferdinand (changes requested):

PR #141, opened 2022-11-23
remove type parameter from arithmetic domain

jan-ferdinand:
We only ever need arithmetic domains over `BFieldElement`s. Hardcoding it makes the code less general, but also less confusing, especially with respect to interpolation and evaluation over arithmetic domains.


sshine (approved):

aszepieniec (approved):
LGTM. Mergit.

PR #144, opened 2022-11-28
Master Tables

### LINKED ISSUE #11: Measure and reduce memory consumption
(Originally posted by @sshine.)

- [ ] Create an overview of the most memory-wasteful areas of the code.
- [ ] Avoid generating entire co-set domains
- [ ] Avoid cloning table collection data

Ultimately, it would be neat to be able to do both padding and extending in-memory; padding is already done in-place, but extending isn't because of the shift from `BFieldElement` to `XFieldElement`. Completely shifting to `XFieldElement` for everything has the performance drawback being that it inflates the proof with zeroes. (It's also a bit obscure.)

We might want to think of a way to pre-allocate space for `XFieldElement`s in the original execution trace, even though it's technically `BFieldElement`s.

(Originally issue number 33 in the internal issue tracker.)

### LINKED ISSUE #119: distinguish AIR constraints for base tables and extension tables
Multiplying two `XFieldElement`s requires 12 multiplications of `BFieldElement`s. AIR constraints that only pertain to a table's base columns do not need to have type `XFieldElement`, but for reasons of convenience, they currently do.

Refactor the code, and possibly the involved traits such as `Evaluable`, to distinguish between constraints for the base table and constraints for the extension table.

### LINKED ISSUE #120: transpose codewords only once when proving
When evaluating the AIR constraints to compute the quotient codewords, the STARK prover currently treats each table separately. For each table, the table's corresponding trace codewords over the arithmetic domain (see #114) are computed, tansposed, and used as evaluation points on the table's AIR constraints.

Refactor the code such that concatenating all trace codewords and transposing the result can be used for all table's constraints. This probably requires a consistent globally canonical ordering of all tables,[^1] globally unique column indices derived from that ordering, and indexing into the transposed codewords using these indices.

[^1]: We probably have that already. To the best of my knowledge, we currently don't rely on its existence.

### LINKED ISSUE #139: present all table data using one “master table”
Triton VM currently has 7 different tables. Each table serves some specific purpose, _e.g._, guaranteeing consistency of the RAM. For separation of logic, this choice is excellent. However, the data is also separated, leading to a few drawbacks:
- multiple indexing schemes (individual-table-based vs master-table-based[^0])
- less direct access to actual data, leading to:
- difficulties in parallelization,
- difficulties in transposing all data, _et cetera_.

I suggest to move the data contained in the individual tables to one big Master Table. This table forms a contiguous region of memory, which is greatly beneficial when moving data to, say, a GPU.[^1] To the extend possible, implementations for [indexing traits](https://doc.rust-lang.org/std/ops/trait.Index.html) _et cetera_ should be provided.

It probably makes sense to split the Master Table into two parts, one of which explicitly refers to base columns (and `BFieldElement`s) and the other to extension columns (and `XFieldElement`s). This allows using compute-optimal arithmetic operations, for example when evaluating constraints.

The current, individual table's column indices must be changed to be globally unique.[^2] A necessary requirement for this is a globally consistent ordering of the tables – something we probably already have, but hopefully never enforce.

Credit & thanks to @Sword-Smith, @aszepieniec, and @einar-triton for the discussion, insights, and ideas.

[^0]: Even though there currently is no “Master Table,” building the base and extension trees _already_ requires essentially the proposed layout, and re-arranges (or points to, using non-trivial index manipulation) the corresponding data.
[^1]: One way to achieve this contiguity is by having all data live inside one big `Vec<_>`.
[^2]: This _might_ be made nice by implementing indexing traits.

### LINKED ISSUE #148: set padded height correctly in the claim
As demonstrated by #147, setting the padded height to 0 in a `Claim` causes some issues. Concretely, the length of the FRI domain indirectly depends on the padded height. The verifier, upon seeing a claimed padded height of 0, reads the actual value from the proof stream instead. However, the prover simply uses the 0, and derives an incorrect FRI domain. Hence, proving & verification does not work in these cases.

Since both the prover and the verifier need to have the actual padded height available, I suggest to disallow the placeholder value 0 and always use the correct padded height in a claim. Furthermore, it is always the prover (never the verifier) claiming something. Computing the correct padded height is trivial for the prover.

### LINKED ISSUE #155: enforce RAM initializating to all zero
In our current design, we can't enforce that the RAM is initialized to all zeros. As discussed with @aszepieniec, the following changes are the least invasive way of achieving this stated goal.

1. Upon _encountering_ (as opposed to _executing_) either instruction `read_mem` or `write_mem`, the processor's registers `RAMP` and `RAMV` get updated.
2. In the RAM Table, whenever a new contiguous `RAMP` region starts, the `RAMV` is 0. This includes an initial constraint on `RAMV` being 0.

I'm happy to provide assistance and more details to anyone considering tackling this issue. It's probably best to wait until #139 is done, _i.e._, #144 is merged.

jan-ferdinand:
A Master Table is, in some sense, a top-level table of Triton VM. It contains all the data but little logic beyond bookkeeping and presenting the data in a useful way. Conversely, the individual tables contain no data but all of the respective logic. Master Tables are responsible for managing the individual tables and for presenting the right data to the right tables, serving as a clean interface between the VM and the individual tables.

As a mental model, it is perfectly fine to think of the data for the individual tables as completely separate from each other. Only the cross-table argument links all tables together.

Conceptually, there are three Master Tables: the Master Base Table, the Master Extension Table, and the Master Quotient Table. The lifecycle of the Master Tables is as follows:

1. The Master Base Table is instantiated and filled using the Algebraic Execution Trace. This is the first time a Master Base Table is instantiated. It is in column-major form.
2. The Master Base Table is padded using logic from the individual tables.
3. The still-empty entries in the Master Base Table are filled with random elements. This step is also known as “trace randomization.”
4. Each column of the Master Base Table is low-degree extended. The result is the Master Base Table over the FRI domain. This is the second and last time a Master Base Table is instantiated. It is in row-major form.
5. The Master Base Table and the Master Base Table over the FRI domain are used to derive the Master Extension Table using logic from the individual tables. This is the first time a Master Extension Table is instantiated. It is in column-major form.
6. The Master Extension Table is trace-randomized.
7. Each column of the Master Extension Table is low-degree extended. The result is the Master Extension Table over the FRI domain. This is the second and last time a Master Extension Table is instantiated. It is in row-major form.
8. Using the Master Base Table over the FRI domain and the Master Extension Table over the FRI domain, the Quotient Master Table is derived using the AIR. Each individual table defines that part of the AIR that is relevant to it.

The following points are of note:
- The Master Extension Table's rightmost columns are the randomizer codewords. These are
   necessary for zero-knowledge.
- The terminal quotient of the cross-table argument, which links the individual tables together,
   is also stored in the Master Quotient Table. Even though the cross-table argument is not
   a table, it does define part of the AIR. Hence, the cross-table argument does not contribute
   to padding or extending the Master Tables, but is incorporated when deriving the Master
   Qoutient Table.
- For better performance, it is possible to derive the Master Quotient Table (step 8) from the
   Master Base Table and Master Extension Table over a smaller domain than the FRI domain –
   the “quotient domain.” The quotient domain is a subset of the FRI domain. This
   performance improvement changes nothing conceptually.

---

- [x] one place to store data for all tables – “master base table”
- [x] one place to store all extension columns, which link individual tables – “master extension table”
- [x] global indexing into master tables
- [x] fill in trace for individual tables
- [x] pad individual tables
- [x] extend individual tables
- [x] remove old logic for trace filling, padding, and extending
- [x] add randomizer polynomials to master extension table, where they can live in their native field
- [x] trace-randomize master tables
- [x] perform low-degree extension in parallel and as in-place as currently possible
- [x] build Merkle trees from master tables
- [x] get quotient degree bounds of individual tables
- [x] compute quotient codewords using master tables
- [x] avoid recomputing zerofiers by building quotients by type, not table
- [x] one place to store all quotients – “master quotient table”
	- [x] including `GrandCrossTableArg` as a first-class citizen
- [x] adapt multicircuits to use master tables
	- [x] adapt code generator correspondingly
	- [x] remove multivariate polynomials from multicircuits
- [x] use [type layout](https://doc.rust-lang.org/stable/reference/type-layout.html#primitive-representations) `#[repr(usize)]` for `*Table*Column` to safely allow indexing with compile-time known indices
- [x] update specification with respect to
	- [x] remove unenforceable Ram Table constraints (#155)
	- [x] add new column “previous instruction” and the necessary constraints surrounding it
	- [x] don't require the Ram Table to be sorted by `ramp`, which is unenforceable anyway
- [x] documentation, tests, cleanup, _et cetera_

---

Fix #11, fix #119, fix #120, fix #139, fix #148, fix #155.


triton-vm/benches/verify_halt.rs
```diff
@@ -22,32 +28,43 @@ fn verify_halt(criterion: &mut Criterion) {
...
     let filename = "halt.tsp";
-    let proof = if proof_file_exists(filename) {
-        match load_proof(filename) {
+    let (proof, stark) = if proof_file_exists(filename) {
+        let proof = match load_proof(filename) {
             Ok(p) => p,
             Err(e) => panic!("Could not load proof from disk: {:?}", e),
-        }
+        };
+        let padded_height = proof.0[1].value() as usize; // todo: Allow creating claim from proof?
```
sshine, line 40:
```suggestion
        let padded_height = proof.padded_height();
```

jan-ferdinand: Ah nice!

jan-ferdinand: Which branch does that live on, though?

PR #146, opened 2022-12-02
Add processor frequency to profiler

AlexanderLemmens:
fix some linting warnings

### DIFF
--- a/triton-profiler/src/triton_profiler.rs
+++ b/triton-profiler/src/triton_profiler.rs
@@ -383,8 +383,6 @@ impl Report {
 
 impl Display for Report {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        writeln!(f, "clock frquency is {} hertz", self.clock_frequency)?;
-
         let max_name_width = self
             .tasks
             .iter()
@@ -456,6 +454,9 @@ impl Display for Report {
                 task_name_colored, padding, task_time_colored, relative_time_string_colored,
             ))?;
         }
+
+        writeln!(f)?;
+        writeln!(f, "Clock frequency is {} Hz", self.clock_frequency)?;
 
         Ok(())
     }
--- a/triton-profiler/src/triton_profiler.rs
+++ b/triton-profiler/src/triton_profiler.rs
@@ -454,7 +454,7 @@ impl Display for Report {
                 writeln!(f)?;
                 writeln!(
                     f,
-                    "Clock frequency is {} Hz ({} clock cycles padded / {} ms)",
+                    "Clock frequency is {} Hz ({} clock cycles / {} ms)",
                     freq, cycle_count, total_time
                 )?;
             }
@@ -467,7 +467,7 @@ impl Display for Report {
                 writeln!(f)?;
                 writeln!(
                     f,
-                    "Optimal clock frequency is {} Hz ({} clock cycles padded / {} ms)",
+                    "Optimal clock frequency is {} Hz ({} padded height / {} ms)",
                     optimal_freq, padded_height, total_time
                 )?;
             }
--- a/triton-profiler/src/triton_profiler.rs
+++ b/triton-profiler/src/triton_profiler.rs
@@ -274,7 +274,10 @@ impl TritonProfiler {
         );
         self.total_time = self.timer.elapsed();
         if let Some(cycle_count) = cycle_count {
-            self.clock_frequency = cycle_count / self.total_time.as_secs() as usize;
+            let total_time = self.total_time.as_secs() as usize;
+            if total_time != 0 {
+                self.clock_frequency = cycle_count / total_time;
+            }
         }
     }
 }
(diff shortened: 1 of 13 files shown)


triton-profiler/src/triton_profiler.rs
```diff
@@ -75,9 +77,8 @@ impl TritonProfiler {
             self.stack.is_empty(),
             "Cannot generate report before stack is empty."
         );
-        assert_ne!(
-            Duration::ZERO,
-            self.total_time,
+        assert!(
+            self.total_time != Duration::ZERO,
```
jan-ferdinand, line 81: This is probably not a desirable change.

triton-profiler/src/triton_profiler.rs
```diff
@@ -176,7 +178,7 @@ impl TritonProfiler {
         });
 
         if std::env::var(GET_PROFILE_OUTPUT_AS_YOU_GO_ENV_VAR_NAME).is_ok() {
-            println!("start: {name}");
+            println!("stop: {name}; took {now:.2?}");
```
jan-ferdinand, line 181:
```suggestion
            println!("start: {name}");
```

triton-profiler/src/triton_profiler.rs
```diff
@@ -232,7 +234,7 @@ impl TritonProfiler {
         self.profile[index].time = duration;
 
         if std::env::var(GET_PROFILE_OUTPUT_AS_YOU_GO_ENV_VAR_NAME).is_ok() {
-            println!("stop:  {name} – took {duration:.2?}");
+            println!("stop: {name}");
```
jan-ferdinand, line 237:
```suggestion
            println!("stop:  {name} – took {duration:.2?}");
```

triton-vm/Cargo.toml
```diff
@@ -27,7 +27,7 @@ default-features = false
 
 [dependencies]
 twenty-first = { version = "0.7.2" }
-triton-profiler = { version = "0.7.2" }
+triton-profiler = { version = "0.7.2", path = "../triton-profiler" }
```
jan-ferdinand, line 30: We probably want users to use the crates.io package, not from the local path.

triton-vm/src/proof_item.rs
```diff
@@ -119,8 +119,7 @@ pub enum ProofItem {
 impl MayBeUncast for ProofItem {
     fn uncast(&self) -> Vec<BFieldElement> {
         if let Self::Uncast(vector) = self {
-            let mut str = vec![];
-            str.push(BFieldElement::new(vector.len().try_into().unwrap()));
+            let mut str = vec![BFieldElement::new(vector.len().try_into().unwrap())];
```
jan-ferdinand, line 122: This is probably an improvement but doesn't really belong to the current scope. I have a slight (but not strong) preference to do it in a separate cleanup commit.

triton-profiler/src/triton_profiler.rs
```diff
@@ -264,13 +266,16 @@ impl TritonProfiler {
...
-    pub fn finish(&mut self) {
+    pub fn finish(&mut self, cycle_count: Option<usize>) {
         assert!(!self.profile.is_empty(), "Nothing to finish.");
         assert!(
             self.stack.is_empty(),
             "Cannot finish before stack is empty."
         );
         self.total_time = self.timer.elapsed();
+        if let Some(cycle_count) = cycle_count {
+            self.clock_frequency = cycle_count / self.total_time.as_secs() as usize;
```
jan-ferdinand, line 277: Be aware of division-by-zero errors here. This is also the reason the test suite failed.

jan-ferdinand (changes requested):
Nice work! After having addressed the comments, I suggest to
1. squash the commits in this branch,
2. supply a descriptive commit message, and
3. rebase & fast-forward merge.

triton-profiler/src/triton_profiler.rs
sshine, line 81:
```suggestion
        assert_ne!(
            self.total_time,
            Duration::ZERO,
```

jan-ferdinand: We usually do `assert_eq!(desired, actual)`. I'm not quite sure how that translates to `assert_ne!(…, …)` but usually go with `assert_ne!(undesired, actual)`.

aszepieniec:
About merging finish and report: you might want to keep the option of separating them. For instance, if there is a benchmark in which not everything is being profiled. Then you want to tell the profiler, "stop tracking time now" but without spending cycles to prepare the report.

jan-ferdinand:
> without spending cycles to prepare the report

Isn't it irrelevant when to spend the cycles if tracking has stopped?

aszepieniec:
> > without spending cycles to prepare the report
> 
> Isn't it irrelevant when to spend the cycles if tracking has stopped?

Yes, we don't care (too much) about spending cycles after tracking has stopped. But one might want to profile a follow-up task or mother task with a separate profiler object, or even another benchmarker altogether. In this case, cycle-tracking is still on-going after the first `finish`.

It is a rather niche use case, but since the functionality is already there I just want to point out what we are making impossible without reverting changes.

triton-vm/Cargo.toml
sshine: Correct. I'll change that now upon release; we needed to use the local one while developing.

PR #147, opened 2022-12-06
Having a padded_height of 0 fails because Stark::new() expects non-zero

sshine:
Co-authored-by: Ferdinand <ferdinand@neptune.cash>


jan-ferdinand:
It was necessary to resolve this issue in #144.

PR #161, opened 2023-01-05
Tip5 Hash Function

aszepieniec:
This TIP specifies a new arithmetization-oriented hash function and proposes to replace Rescue-Prime with it.

See [here](https://github.com/TritonVM/triton-vm/blob/asz/tip5/tips/tip-0005/tip-0005.md) for the markdown document.


jan-ferdinand:
@aszepieniec, is this PR ready to be merged or is there something more to be included?

aszepieniec:
The new and improved MDS matrix multiplication is work-in-progress. It matters because it determines different test vectors.

After that, the big task is obviously to use Tip5 and program its arithmetization into the AIR .... but I guess that can be a different pull request?

jan-ferdinand:
I suggest we make it a different PR, yes. :relieved:

PR #159, opened 2022-12-29
U32 Table

jan-ferdinand:
- [ ] spec
    - [x] u32 table
    - [x] instructions
    - [x] pseudo instructions
    - [x] processor table
    - [x] instruction specific transition constraints
    - [x] AET relation diagram
    - [ ] cheat sheet
- [ ] code   
    - [ ] add new instructions
    - [ ] step mut
    - [x] u32 table scaffolding
    - [ ] u32 fill
    - [ ] u32 pad
    - [ ] u32 extend
    - [ ] processor perm arg
    - [ ] processor instruction transition constraints
    - [ ] grand cross-table arg
    - [ ] rework pseudo instructions (see updated spec)
- [ ] improvements
    - [ ] use IB_3 through IB_5 in U32 Table, get low-degree de-selector

One possible way to address #8. Does not necessarily close that issue.

PR #162, opened 2023-01-11
nom-parser

sshine:
The existing parser has a number of bugs and oddities.

This PR re-implements the parser using [nom](https://github.com/rust-bakery/nom).

Main changes:
- Better, pretty syntax error messages
- Fewer errors handling comments, whitespace, constants, labels, unicode
- Removed support for pseudo-instructions in favor of `[foo]` as syntax sugar for `call foo`

Since [tasm-lib](https://github.com/TritonVM/tasm-lib) depends on a number of these pseudo-instructions, and since #159 converts some of these to real instructions, replacing the parser at `Program::from_code()` currently has too many dependencies.

Tasks that can happen after this parser is merged:
- Before #159 is merged: Update nom parser to handle instructions introduced in #159.
- After #159 is merged: Replace any pseudo-instructions that hasn't become an instruction with a tasm-lib snippet. Handle cost of `call`-`return` later by automated inlining.


sshine:
As of bfe53b8b1a87ee5081dcefb03728946c806c2c3c, `nom-parser` now supports

- `lt`, `and`, `xor`, `log_2_floor`, `pow`, `div`

and it removed support for `lsb` which has become a pseudo-instruction.

PR #170, opened 2023-01-23
add instruction `swap_digest`

jan-ferdinand:
The new instruction `swap_digest` swaps the five top-most elements with the next five elements.


jan-ferdinand:
It is as of yet unknown whether `swap_digest` widens a bottleneck – converting the PR to draft status.

PR #171, opened 2023-01-27
replace instruction table by lookup argument

jan-ferdinand:
Replace the Instruction Table with a Lookup Argument, as described [here](https://ia.cr/2023/107). This decreases the total number of columns by 5, the total number of constraints by 9, and is expected to decrease the number of rows. The last property depends on the concrete program and its input. However, in the worst case, the number of rows remains unchanged.


triton-vm/src/vm.rs
```diff
@@ -39,15 +39,22 @@ pub fn simulate(
...
+
     // record initial state
     aet.processor_trace
         .push_row(state.to_processor_row().view())
         .expect("shapes must be identical");
 
     let mut stdout = vec![];
     while !state.is_complete() {
+        // If the instruction pointer is out of bounds, the corresponding error handling happens
+        // in the state transition function `step_mut`.
```
sshine, line 54: I might like the error to occur here, too.

triton-vm/src/vm.rs
```diff
@@ -136,18 +157,19 @@ pub struct AlgebraicExecutionTrace {
     pub u32_entries: Vec<(Instruction, BFieldElement, BFieldElement)>,
 }
 
-impl Default for AlgebraicExecutionTrace {
-    fn default() -> Self {
+impl AlgebraicExecutionTrace {
+    pub fn new(program: Program) -> Self {
+        let instruction_multiplicities = vec![0_u32; program.len()];
```
sshine, line 162: I know I made `.len()` on `Program`, but in retrospect, this is ambiguous to me:

A program's length could either be the number of instructions, or the number of b-words.

It is actually the number of b-words here, but perhaps `.len()` could be more clear.

sshine (approved):

triton-vm/src/vm.rs
jan-ferdinand: We could have a method 

```rust
pub fn number_of_instructions(&self) -> usize {
    self.clone().into_iter().count()
}
```

as addition to `.len()`. I decided against that since it would not be used anywhere. I tried to clarify the behavior of `.len()` by introducing the doc-string. Thoughts?

sshine: We don't have to have a function that never gets called.

I suppose what I'm suggesting is to rename `.len()` into something unambiguous, like `.len_bwords()`.

jan-ferdinand: That makes sense and is in line with the naming scheme in `triton-opcodes`. I'll refactor accordingly. :+1:

triton-vm/src/table/master_table.rs
```diff
@@ -327,11 +320,12 @@ impl MasterBaseTable {
         roundup_npo2(max_height as u64) as usize
     }
 
-    pub fn instruction_table_length(
-        aet: &AlgebraicExecutionTrace,
-        program: &[BFieldElement],
-    ) -> usize {
-        program.len() + aet.processor_trace.nrows()
+    pub fn program_length(aet: &AlgebraicExecutionTrace) -> usize {
```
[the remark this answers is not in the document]
aszepieniec: Pattern matching on the context suggests `master_base_table.program_table_length()`.

triton-vm/src/vm.rs
aszepieniec: Is there ever a context where we need the number of instructions without counting the arguments?

aszepieniec (approved):
LGTM

triton-vm/src/vm.rs
jan-ferdinand: Not that I know of.

aszepieniec: I guess I'm a little confused about what ambiguity this renaming is supposed to eliminate. What other thing could program length refer to?

PR #177, opened 2023-02-03
replace involved argument for clock jump differences with 3 Lookup Args

jan-ferdinand:


aszepieniec (changes requested):
It looks like there are 3 separate Lookup Arguments. How about merging the servers? That shouldn't be too difficult.

The next step is to merge the clients also. I believe that's possible but will require some more engineering work.

jan-ferdinand:
> How about merging the servers? That shouldn't be too difficult.

I agree: the engineering effort is pretty low, there's no real reason to separate the arguments, and it saves a few columns.

> The next step is to merge the clients also.

I disagree: the red wedding should happen before we take a closer look at whether merging the clients is worth it.

aszepieniec (approved):
LGTM

sshine (approved):

PR #179, opened 2023-02-09
change behavior of instructions `read_mem` and `write_mem`

jan-ferdinand:
Up until now, both instructions did not modify the operational stack's length. The new behavior is:

- `read_mem` uses stack element `ST0` as the memory pointer `ramp` and pushes the memory value `ramv` to the stack, increasing its size.
- `write_mem` (still) uses stack element `ST1` as the memory pointer `ramp` and writes stack element `ST0` to RAM, popping it from the stack in the process.


triton-vm/Cargo.toml
```diff
@@ -30,7 +30,7 @@ version = "1"
 default-features = false
 
 [dependencies]
-triton-opcodes = "0.14"
+triton-opcodes = {path = "../triton-opcodes"}
```
sshine, line 33: Remember to update after releasing triton-opcodes.

sshine (approved):

Sword-Smith (approved):

Sword-Smith:
Seems like most test programs get a bit shorter. Nice!

PR #183, opened 2023-02-24
remove duplicate nodes from constraint circuits

jan-ferdinand:
This is achieved by doing constant folding in the monad, not in the expression.


Sword-Smith (approved):
Added a few commits to delete commented-out code and fixed a test.

It was very nice fixing this bug with you, @jan-ferdinand. This has been a pebble in my shoe for months.

The constant folding function could easily be made a lot faster by adding a new field to the monad context, but since it seems fast enough now, there's no good reason to speed that up, I think.

Sword-Smith:
The only criticism I have of this PR as it looks now is that there's more boilerplate code in the functions that define the circuits. They all end with incantations similar to this:
```rust
let mut constraints = [
            osp_increases_by_1_or_does_not_change,
            osp_increases_by_1_or_osv_does_not_change_or_shrink_stack,
            rppa_updates_correctly,
            log_derivative_updates_correctly,
        ];

        ConstraintCircuitMonad::constant_folding(&mut constraints);
        constraints.map(|circuit| circuit.consume()).to_vec()
```

jan-ferdinand:
I have thought the same thing but considered it benign enough to delegate the improvement slumbering here to future contributors (which might be future-us).

PR #184, opened 2023-03-02
update to twenty-first v0.18

jan-ferdinand:


triton-vm/src/proof_stream.rs
```diff
@@ -109,12 +108,26 @@ where
         Ok(ProofStream {
             items,
             items_index: 0,
-            _hasher: PhantomData,
+            sponge_state: H::init(),
         })
     }
 
+    fn encode_and_zero_pad_item(item: &Item) -> Vec<BFieldElement> {
```
aszepieniec, line 115: There is no appending of `1` going on before appending the `0`s. Shouldn't that be there?

aszepieniec:
I wonder if now is a good time to mark some `ProofItems` as not for hashing, for the purpose of Fiat-Shamir. This applies to all `ProofItem`s that are computationally uniquely determined by another `ProofItem`. For example, a Merkle authentication path does not need to be hashed, whereas a Merkle root does.

jan-ferdinand:
It might be a good time, yes.

triton-vm/src/proof_stream.rs
jan-ferdinand: The idea is to simplify Fiat-Shamir for the recursive verifier, which always absorbs all currently available elements and doesn't keep “stray tails” around until the next RATE many elements are reached. If need be, the difference is filled with 0s.

aszepieniec: I think the `1` should be inserted there. Otherwise you might have two different items that generate the same encoded and zero-padded string, and thus the same Fiat-Shamir challenge.

jan-ferdinand: This also simplifies the padding for the variable length hashing greatly, because that is always the vector [1, 0, …, 0]. However, that padding is not yet happening anywhere and must be added.

jan-ferdinand: > I think the `1` should be inserted there. Otherwise you might have two different items that generate the same encoded and zero-padded string, and thus the same Fiat-Shamir challenge.

That also works; whenever any element is absorbed, we act as if it might be the last one, and are thus ready for sampling indices or scalars.

jan-ferdinand: However, it might be more efficient to get an explicit call like “.fiat_shamir()” which appends the padding for variable length input, and only include the 1 then. This means that the recursive verifier does not have to spend cycles on figuring out what the correct amount of padding is.

aszepieniec: As I see it there are three options:

 1. Append `1` and `0`s after enqueuing or dequeuing any `ProofItem`, thereby being prepared for Fiat-Shamir even if that happens not to be the next step.
 2. Store the elements in a separate queue and whenever the queue has `rate`-many elements or more, absorb them away. When Fiat-Shamir happens, append the `1` and `0`s and flush the queue by absorbing all of it before squeezing.
 3. Pad with zeros to absorb all `ProofItems` as they are enqueued or dequeued, and just squeeze to get the Fiat-Shamir response. In this approach, it is immaterial whether a `1` was appended prior to squeezing out the Fiat-Shamir challenge. This approach can only be secure under certain assumptions on the serializer for `ProofItem`s.

Imposing constraints for security on `ProofItem`s' serializers seems sloppy and prone to error to me, so (3) is out. Between (1) and (2) I think (2) has fewer invocations of the hash function but at the cost of more bookkeeping. It's not obvious to me that (1) results in a worse recufier cycle count; I think that argument might go either way. For now, I would suggest the simpler option which in my eyes is (1).

aszepieniec:
> Have you already spent brain cycles on potential realizations of this idea or should I start from scratch?

I would add a `bool` to `enqueue` and `dequeue` that activates the sponge update. If the sequence of true/false in the prover does not match that in the verifier, then prover and verifier cannot possibly agree on the Fiat-Shamir challenges. So the proof system's completeness enforces the consistent marking.

triton-vm/src/proof_stream.rs
jan-ferdinand: Option 4: same as option 3, but with an explicit absorption of vector [1, 0, …, 0] to mark the end of en- or de-queueing.

> it is immaterial whether a 1 was appended prior to squeezing

I don't see why this holds.

aszepieniec: If I understand option 4 correctly:

 - The string `[0]` will be padded with zeros to `[0,0,0,0,0,0,0,0,0,0]`, then absorbed setting the sponge state to X. Then calling Fiat-Shamir will absorb `[1,0,0,0,0,0,0,0,0,0]` followed by a squeeze resulting in output Y.
 - The string `[0, 0]` will be padded with zeros to `[0,0,0,0,0,0,0,0,0,0]`, then absorbed setting the sponge state to X. Then calling Fiat-Shamir will absorb `[1,0,0,0,0,0,0,0,0,0]` followed by a squeeze resulting in output Y.

There are therefore two distinct strings that generate the same Fiat-Shamir challenge. This is either insecure, or else it is secure due to some analysis of the possible sequences of field elements that could be enqueued or dequeued.

jan-ferdinand: Ah, true. Alright, let's roll with option 1 and see if it requires further optimization down the line.

aszepieniec (approved):

PR #185, opened 2023-03-13
parse instructions `dup` and `swap` as taking arguments

jan-ferdinand:
Requires `dup` and `swap` instructions to be written as `dup n` (respectively `swap n`) instead of `dupn` (respectively `swapn`). This reflects the internals of Triton VM.


Sword-Smith:
Looks benign to me. But what's the reason you want to make this change?

Sword-Smith (approved):

PR #187, opened 2023-03-15
add instruction `pop_count`

jan-ferdinand:
`pop_count` is a u32 instruction that computes the hamming weight of its argument. The argument is the stack's top-most element, which is replaced with the result. As for all u32 instructions, the VM crashes if the argument is not a u32.


Sword-Smith:
Excited about this :))

jan-ferdinand:
The current naming scheme for [all our instructions](https://triton-vm.org/spec/instructions.html) is to separate distinct words with an underscore. I see the word “count” as distinct and interpret “pop” as the shorthand for “population.”

An alternative name could be `hamming_weight`, which is long but not longer than `divine_sibling`.

PR #190, opened 2023-04-19
implement DEEP-ALI

jan-ferdinand:


triton-vm/src/stark.rs
```diff
@@ -182,7 +186,8 @@ impl Stark {
         prof_start!(maybe_profiler, "ext tables");
         prof_start!(maybe_profiler, "LDE");
         master_ext_table.randomize_trace();
-        let fri_domain_ext_master_table = master_ext_table.to_fri_domain_table();
+        let (fri_domain_ext_master_table, ext_interpolation_polys) =
+            master_ext_table.to_fri_domain_table();
```
aszepieniec, line 190: I don't think we need to keep track of the polynomials if we have barycentric evaluation. File under: optimizable-if-bottleneck.

triton-vm/src/stark.rs
```diff
@@ -192,10 +197,38 @@ impl Stark {
...
         prof_stop!(maybe_profiler, "ext tables");
 
-        prof_start!(maybe_profiler, "quotient degree bounds");
-        let quotient_degree_bounds =
-            all_quotient_degree_bounds(self.interpolant_degree, master_base_table.padded_height);
-        prof_stop!(maybe_profiler, "quotient degree bounds");
+        prof_start!(maybe_profiler, "out-of-domain rows");
+        let trace_domain_generator = derive_domain_generator(padded_height.value());
+        let out_of_domain_value_curr_row = proof_stream.sample_scalars(1)[0];
+        let out_of_domain_value_next_row = trace_domain_generator * out_of_domain_value_curr_row;
```
aszepieniec, line 203: I would suggest to use "point" instead of "value" because "value" implies (to my ears at least) that it's the output of some evaluation. Here's it's the opposite: it's the *input*!

triton-vm/src/stark.rs
```diff
@@ -220,108 +253,171 @@ impl Stark {
...
+            base_and_ext_interpolation_poly.evaluate(&out_of_domain_value_next_row);
+        proof_stream.enqueue(&ProofItem::OutOfDomainElement(
+            out_of_domain_next_row_base_and_ext_element,
+        ));
+        let base_and_ext_next_row_deep_quotient = Self::deep_quotient(
+            &base_and_ext_codeword,
+            quotient_domain,
+            out_of_domain_value_next_row,
+            out_of_domain_next_row_base_and_ext_element,
         );
```
aszepieniec, line 315: I would prefer to avoid overloading the term "quotient" because then there will be quotient quotients.

triton-vm/src/stark.rs
```diff
@@ -528,38 +588,138 @@ impl Stark {
...
+        let out_of_domain_curr_base_row = Array1::from(out_of_domain_curr_base_row);
+        let out_of_domain_curr_ext_row = Array1::from(out_of_domain_curr_ext_row);
+        let out_of_domain_next_base_row = Array1::from(out_of_domain_next_base_row);
+        let out_of_domain_next_ext_row = Array1::from(out_of_domain_next_ext_row);
+        prof_stop!(maybe_profiler, "get out-of-domain index and rows");
+
+        prof_start!(maybe_profiler, "out-of-domain quotient element");
+        prof_start!(maybe_profiler, "zerofiers");
+        let one = BFieldElement::one();
+        let trace_domain_generator = derive_domain_generator(padded_height as u64);
```
aszepieniec, line 610:
```
let trace_domain_generator = derive_domain_generator(padded_height as u64);
```
Did I read that line twice?

triton-vm/src/stark.rs
```diff
@@ -602,210 +762,127 @@ impl Stark {
...
-        let revealed_combination_leafs =
+        prof_start!(maybe_profiler, "Merkle verify (combined quotient)");
+        let revealed_quotient_elements =
             proof_stream.dequeue()?.as_revealed_combination_elements()?;
-        let revealed_combination_digests = revealed_combination_leafs
+        let revealed_quotient_digests = revealed_quotient_elements
             .par_iter()
-            .map(StarkHasher::hash)
+            .map(|&x| x.into())
             .collect::<Vec<_>>();
```
aszepieniec, line 779: I'm not sure what these three lines are supposed to do.
```
let revealed_quotient_digests = revealed_quotient_elements
            .par_iter()
            .map(|&x| x.into())
```
Maybe add an explicit type to clarify?

aszepieniec (approved):
LGTM

triton-vm/src/stark.rs
jan-ferdinand: How do we get the out-of-domain evaluation of the AET if we don't have the columns' interpolation polynomials (in coefficient form)?

triton-vm/src/stark.rs
jan-ferdinand: I have a added the following comment instead:
```
// Interpret the leaves, which are XFieldElements, as Digests, without hashing.
```

Does that clarify the operation sufficiently?

PR #192, opened 2023-04-26
avoid unnecessary hashing in the Fiat-Shamir heuristic

jan-ferdinand:
In particular, anything that has already been committed to does not need to be committed to again. Similarly, if no more randomness is sampled, changing the Sponge's state is unnecessary.


aszepieniec (approved):
LGTM

Issue #205, opened 2023-06-16
Completeness bug involving `lt`


Sword-Smith:
I found a new completeness bug in Triton VM. I spent a few hours attempting to narrow down the bug, but so far, I've been unable to identify it fully. However: When I change the line below from `dup 2 lt` to `dup 2 eq push 0 eq` (inside the "safe push" subroutine), which should be equivalent, it succeeds. So we might be looking at another problem with `lt`. I haven't been able to find that `lt` problem in isolation though.

```
push 0
push 3
push 1
push 457470286889025784
push 4071246825597671119
push 17834064596403781463
push 17484910066710486708
push 6700794775299091393
push 6
push 02628975953172153832
write_mem
pop
push 10
push 01807330184488272967
write_mem
pop
push 12
push 06595477061838874830
write_mem
pop
push 1
push 2
write_mem
pop
push 11
push 10897391716490043893
write_mem
pop
push 7
push 01838589939278841373
write_mem
pop
push 8
push 05057320540678713304
write_mem
pop
push 4
push 00880730500905369322
write_mem
pop
push 5
push 06845409670928290394
write_mem
pop
push 3
push 04594396536654736100
write_mem
pop
push 2
push 64
write_mem
pop
push 9
push 05415221245149797169
write_mem
pop
push 0
push 323
write_mem
pop

            call tasm_mmr_calculate_new_peaks_from_append_safe
            halt

            
                // BEFORE: _ old_leaf_count_hi old_leaf_count_lo *peaks [digests (new_leaf)]
                // AFTER: _ *new_peaks *auth_path
                tasm_mmr_calculate_new_peaks_from_append_safe:
                    dup 5 dup 5 dup 5 dup 5 dup 5 dup 5
                    call tasm_list_safe_u32_push_digest
                    pop pop pop pop pop
                    // stack: _ old_leaf_count_hi old_leaf_count_lo *peaks

                    // Create auth_path return value (vector living in RAM)
                    push 64 // All MMR auth paths have capacity for 64 digests
                    call tasm_list_safe_u32_new_digest
                    push 0
                    call tasm_list_safe_u32_set_length_digest
                    // stack: _ old_leaf_count_hi old_leaf_count_lo *peaks *auth_path

                    swap 1
                    // stack: _ old_leaf_count_hi old_leaf_count_lo *auth_path *peaks

                    dup 3 dup 3
[...]

PR #207, opened 2023-06-16
Fix opcode error and add regression test related to `lt` instruction

Sword-Smith:
`lt` decreases the opstack size, and this must be reflected in its opcode. This bug was discovered by attempting to prove/verify all `tasm-lib` snippets.

New version of opcodes must be used for this fix to work. That dependency upgrade is not included in this commit, but the fix to `triton-opcodes` is.


Sword-Smith:
Merging this closes #205

PR #208, opened 2023-06-16
Completeness bug involving squeeze

Sword-Smith:
The VM cannot handle `squeeze` as the 1st hash instruction. Initial constraint 1 of the hash table reports an error on this program: `squeeze halt`. Two programs illustrate this: `SAMPLE_INDICES_USING_SAFE_LISTS` and `test_program_for_squeeze`.


jan-ferdinand:
If the first hashing-related instruction is a Sponge instruction, it has to be `absorb_init`. [The failing initial constraint](https://triton-vm.org/spec/hash-table.html#initial-constraints) performs exactly this check. That said, I see at least three shortcomings identified by reporting this supposed bug:

1. It is impossible that a Sponge's first action is a squeeze, but it should be.
1. The [documentation regarding the hash instructions](https://triton-vm.org/spec/instructions.html#hashing) is not good enough.
1. The ergonomics of the current Sponge instructions are bad.

I will look into improving Triton VM on those points.

PR #216, opened 2023-08-03
non-deterministic initialization of RAM

jan-ferdinand:


triton-vm/src/program.rs
```diff
@@ -385,6 +385,109 @@ impl ProfileLine {
...
+
+impl PublicInput {
+    pub fn new(stream: Vec<BFieldElement>) -> Self {
+        PublicInput { stream }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct SecretInput {
+    pub stream: Vec<BFieldElement>,
```
aszepieniec, line 437: I don't really like the name "stream" because it suggests some kind of logspace computation on a large amount of linearly arrayed data, along with the capacity to compose processes and chain streams. But I'm not sure what alternative to use. Source, perhaps?

aszepieniec: Also, see my earlier comment about splitting up this field into two, one for each type of divination.

aszepieniec (approved):
The constraints look good to me.

I would like to make `SecretInput` more powerful, starting by separating the `stream` field into two separate sources of non-determinism. Also, would be nice to load a `BFieldCodec`-encodable object into initial memory at a given or at the first free address.

Also, can we call "secret input" something else? Its defining quality is not that it is secret, nor that it is a tape to be read from like public input is. Its defining quality is that it uniquely determines an otherwise ill-defined trace. With that description in mind, I suggest "NonDeterminism".

Sword-Smith:
> Also, would be nice to load a BFieldCodec-encodable object into initial memory at a given or at the first free address.

This seems to me like putting too much responsibility on the `run` or `program` interface. Maybe what @aszepieniec suggests fits better as a helper function either here in `triton-vm` or in [`tasm-lib`](https://github.com/TritonVM/tasm-lib)?

Sword-Smith:
> This seems to me like putting too much responsibility on the `run` or `program` interface. Maybe what @aszepieniec suggests fits better as a helper function either here in `triton-vm` or in [`tasm-lib`](https://github.com/TritonVM/tasm-lib)?

I guess a way of constructing initial memory could be added to either this repo or to `tasm-lib`? `InitialMemoryBuilder` could add a `TasmObject` struct to the initial memory and then return the address on which it is being put into memory? Or, rather, return a counter that counts how many of these `TasmObject`s that have been added to memory. This method could e.g. be called `add_to_initial_memory` or something like this. It's worth noting, though, that the `HashMap` of initial memory cannot be known at compile time, and neither can the addresses on which each object is located as those objects are dynamically sized.

PR #221, opened 2023-08-16
Add `Default` derivation to `NonDeterminism`

Sword-Smith:
We invoking the VM we often need an empty non-deterministic input, so a convenience function for this would be nice. As `NonDeterminism` only consists of fields that implement `Default`, `Default` extends naturally to `NonDeterminism`.


jan-ferdinand (approved):
An alternative with fewer characters than `Default::default()` is `[].into()` since there is an `impl From<[u64; 0]> for NonDeterminism`.

PR #224, opened 2023-09-16
add benchmark for longer-running program

jan-ferdinand:


jan-ferdinand:
Closing in favor of #229 because the CI instances reliably keep running out of memory for the bigger benchmark.

Issue #220, opened 2023-08-15
Concatenated labels in TASM


Sword-Smith:
A lot of code in `tasm-lib` creates labels this way:

```rust
let entrypoint = self.entrypoint_name();
let code = format!("
    {entrypoint}:
        call {entrypoint}_inner
        return
        
    {entrypoint}_inner:
        ...
        ...
        return
")
```
Which works fine for string concatenation. But when you do the same with the `triton_asm!` macro, the macro inserts a space between the `entrypoint` string and `_inner` forcing you to pre-declare all labels which, in my opinion, reduces readability.

PR #240, opened 2023-11-17
improve Triton assembly macros

jan-ferdinand:
- allow creating labels like `{label}_tail`, fixing #220
- simplify declaration (of internal relevance only)

It should be possible to make label building more general but I find it difficult to do[^1] so without [metavariable expressions](https://doc.rust-lang.org/beta/unstable-book/language-features/macro-metavar-expr.html), which are currently unstable.

[^1]: probably more a testament to my macro-writing skill than anything else

### DIFF
--- a/triton-vm/src/lib.rs
+++ b/triton-vm/src/lib.rs
@@ -302,14 +302,31 @@ macro_rules! triton_program {
 /// [tasm]: https://triton-vm.org/spec/instructions.html
 #[macro_export]
 macro_rules! triton_asm {
-    (@fmt $fmt:expr, $($args:expr,)*; ) => {
-        format_args!($fmt $(,$args)*).to_string()
-    };
+
     (@fmt $fmt:expr, $($args:expr,)*; $label_declaration:ident: $($tail:tt)*) => {
         $crate::triton_asm!(@fmt
             concat!($fmt, " ", stringify!($label_declaration), ": "), $($args,)*; $($tail)*
         )
     };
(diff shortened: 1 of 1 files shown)


Sword-Smith:
Very nice. Looking forward to this :)

jan-ferdinand:
Experimenting and reading up on the matter, I can't seem to achieve the behavior I want using declarative macros. In particular, I would like to allow building labels like `{expr}ident` but disallow `{expr} ident` – the whitespace in between makes all the difference.[^0] Since macros operate on token trees,[^1] whitespace is not preserved. As a consequence, detecting whether some token is part of a label becomes quite difficult. While it might work for label declarations, uniquely terminated by a colon, a `call` followed by a series of tokens can be ambiguous. Does `call {my_list}push{my_item}` describe a `call` to `{my_list}` followed by instruction `push` with `{my_item}`, or is the intended `call` target `{my_list}push{my_item}`?

A workaround[^2] is to fully declare the desired label within the expression:
```rust
{format!("{my_label}_inner")}: push 0 return
…
call {format!("{my_label}_inner")}
```

[^0]: Successfully matching `{expr} ident` as a label causes issues like `{my_instruction} my_label:` or `call {my_label} halt` being erroneously matched and fused.
[^1]: true for both declarative and procedural macros
[^2]: ugly, but works today

Sword-Smith:
Too bad that this is not possible. Thanks for looking into it though!

jan-ferdinand:
I do plan to add the functionality. By closing this branch, I didn't want to communicate that #220 is off the table. It's just that the approach tried here does not work. :relaxed:

PR #250, opened 2024-03-05
Fix constant folding pbt

Sword-Smith:
Fixes the constant-folding PBT and adds a check to prevent different circuit-builders from being used
when forming binary expressions, as this leads to ID collissions which the constant-folding algorithm
cannot (and should not) handle.

### DIFF
--- a/triton-vm/src/table/constraint_circuit.rs
+++ b/triton-vm/src/table/constraint_circuit.rs
@@ -23,6 +23,8 @@ use num_traits::One;
 use num_traits::Zero;
 use quote::quote;
 use quote::ToTokens;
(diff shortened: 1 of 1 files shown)


Sword-Smith:
The equality check that I added was a reference equality check whereas your is a value equality. The reference equality check is much faster as it doesn't have to traverse the `all_nodes` hash set. Since we're struggling with compilation time in this repo, I think you should favor reference equality.

I'm happy with the state of the constant folding checks that your commit resulted in. As long as we still have (and use) the `evaluate_assert_unique` test helper function, we can be assured that we don't have a degenerate representation of the multi circuit.

PR #270, opened 2024-04-26
chore!: Upgrade to twenty-first 0.42.0-alpha.2

Sword-Smith:


triton-vm/src/lib.rs
```diff
@@ -568,7 +572,11 @@ mod tests {
         #[strategy(arb())] hash_preimage: Digest,
         #[strategy(arb())] some_tie_to_an_outer_context: Digest,
     ) {
-        let hash_digest = hash_preimage.hash::<Tip5>().values();
+        let hash_digest = Tip5::hash_pair(
+            hash_preimage,
+            Digest::new([BFieldElement::zero(); DIGEST_LENGTH]),
```
jan-ferdinand, line 577: Isn't this just
```suggestion
            Digest::default(),
```
?

Alternatively, if it is not:
```suggestion
            Digest::new(bfe_array![0; DIGEST_LENGTH]),
```

Cargo.toml
```diff
@@ -26,7 +26,13 @@ anyhow = "1.0"
...
 assert2 = "0.3"
 colored = "2.1"
-clap = { version = "4", features = ["derive", "cargo", "wrap_help", "unicode", "string"] }
+clap = { version = "4", features = [
+    "derive",
+    "cargo",
+    "wrap_help",
+    "unicode",
+    "string",
+] }
```
jan-ferdinand, line 35: Why the reformat?

triton-vm/src/lib.rs
Sword-Smith: It is just `Digest::default()`. I wasn't sure how comfortable using `default` for this value. I'm OK with it though.

Cargo.toml
Sword-Smith: My IDE did that. I don't have a pronounced opinion.

triton-vm/src/lib.rs
jan-ferdinand: I would say since this is a test, it's fine; If something about `Digest::default()` changes, it will be spotted here immediately.

Sword-Smith:
I force-pushed with the suggested changes. Feel free to merge.

PR #272, opened 2024-04-30
Dot-Step

aszepieniec:
This PR:
 - Introduces two new instructions to the VM, `xxdotstep` and `xbdotstep` which executes one iteration of a typical dot product loop.
 - Adds state evolution logic and constraints for the new instructions.

T~~wo tasks remain before t~~his PR is ready:
 - [x] Update documentation.
 - [x] Fix failing test `constant_folding_processor_table` which detects duplicate nodes.


jan-ferdinand:
I would also ask you to rebase & force-with-lease push to remove “wip” commits. :relieved:

Sword-Smith:
I see you've moved the constant folding. It makes sense to do what you can on insertion, but I wanted to ask if we're still cathing as much as we did previously. I think we are but was wondering if you guys have more insight.

jan-ferdinand:
We've thought it through and concluded that we are catching all the same cases as before. Moving it further ahead in the pipeline serves two benefits (and has no downsides as far as we can see):
- constant folding does not have to be called manually, and
- semantically duplicate nodes are detected earlier, removing the strange behavior that nodes with different IDs evaluate to the same expression.

jan-ferdinand:
Thanks! Good work. :blush:

PR #286, opened 2024-05-23
Use `batch_evaluate`

aszepieniec:
Repo `twenty-first` recently acquired new methods for fast batch evaluation. This PR uses them.

Down the line we should also modify the just-in-time code path to drop the cached polynomials and use fast batch extrapolation (from the same new and improved `twenty-first`) instead. But that's for a separate PR and on a different timeline.


aszepieniec:
Merged via cherry pick. Issue #287 remains open to track the dropping of the cached polynomials.

PR #288, opened 2024-05-23
feat: Introduce instruction `recurse_or_return`

jan-ferdinand:
Instruction `recurse_or_return` behaves – surprise! – either like instruction `recurse` or like instruction `return`. The (deterministic) decision which behavior to exhibit is made at runtime and depends on the current stack as well as the instruction's immediate argument `i`.

Argument `i` indicates one of the stack's top 16 elements. Let the value of the indicated stack element be `a`, and the value of its successor element `b`. In the slightly special case of `i` == 15, the indicated element's successor is taken to be stack element with index 0. If `a` != `b`, then `recurse_or_return` acts like instruction `recurse`, else like `return`.


Sword-Smith:
I'm seeing an 11 % increase in the number of base columns which takes away what I estimate is 30-50 % of the gain that this new instruction yields. How would that number (the increase in the number of base columns) look if you remove the immediate argument to the instruction and only look at stack element 5 and 6 which is what we need for a faster TASM-verifier?

jan-ferdinand:
While I was aware that the additional cost of the new instruction in its current design is quite high, the hope was that raising the degree the constraints are lowered to, which would be possible cheaply with an upcoming update, would remove that additional cost again. However, some quick experiments indicate that this is not so; with a target degree of 8, the new instruction would contribute more than a quarter of the total degree-lowering columns. This is an even higher relative impact: at target degree 4, the current design would be responsible for around 15% of the degree lowering columns.

image

> How would [the increase in the number of base columns] look if you remove the immediate argument to the instruction and only look at stack element 5 and 6 […]?

Probably a lot better. To find out the actual numbers, I need to write the corresponding code.

jan-ferdinand:
> How would [the increase in the number of base columns] look if you remove the immediate argument to the instruction and only look at stack element 5 and 6 […]?

The new instruction `recurse_or_return` without immediate argument introduces an additional 3 base columns in the degree lowering table.

triton-vm/src/op_stack.rs
```diff
@@ -336,6 +338,24 @@ impl Display for OpStackElement {
...
+    }
+}
+
+impl WrappingAdd for OpStackElement {
+    fn wrapping_add(&self, &rhs: &Self) -> Self::Output {
+        let wrapped_sum = (self.index() + rhs.index()) % (OpStackElement::COUNT as u32);
+        wrapped_sum.try_into().unwrap()
+    }
+}
+
```
jan-ferdinand, line 358: This behavior used to be helpful for the instruction's previous iteration. I have left the functionality in but the only argument in favor of this code's existence is “it doesn't hurt to have it.” Let me know if you think it should be removed.

triton-vm/src/table/processor_table.rs
```diff
@@ -1670,6 +1671,52 @@ impl ExtProcessorTable {
...
+        let curr_row = |col: ProcessorBaseTableColumn| {
+            circuit_builder.input(CurrentBaseRow(col.master_base_table_index()))
+        };
+        let next_row = |col: ProcessorBaseTableColumn| {
+            circuit_builder.input(NextBaseRow(col.master_base_table_index()))
+        };
+
+        // Zero if the ST5 equals ST6. One if they are not equal.
+        let st5_eq_st6 = || curr_row(HV0) * (curr_row(ST6) - curr_row(ST5));
+        let st5_neq_st6 = || one() - st5_eq_st6();
```
aszepieniec, line 1687: What's the motivation for defining these as closures rather than constructing nodes and `clone`ing them?

jan-ferdinand: Now, mostly alignment for the code below: I like the looks of

```rust
let maybe_return = vec![
    st5_neq_st6() * curr_row(HV0),
    st5_neq_st6() * (curr_row(ST6) - curr_row(ST5)),
    st5_neq_st6() * (next_row(IP) - curr_row(JSO)),
    st5_neq_st6() * (next_row(JSP) - curr_row(JSP) + one()),
];
```

better than

```rust
let maybe_return = vec![
    st5_neq_st6.clone() * curr_row(HV0),
    st5_neq_st6.clone() * (curr_row(ST6) - curr_row(ST5)),
    st5_neq_st6.clone() * (next_row(IP) - curr_row(JSO)),
    st5_neq_st6 * (next_row(JSP) - curr_row(JSP) + one()),
];
```

because in my opinion, it provides greates visual clarity that the first part of the constraint is indeed identical for all four constraints.

Historically, it was also useful to pass the stack element in question as an argument to the closure.

triton-vm/src/table/processor_table.rs
```diff
@@ -1670,6 +1671,52 @@ impl ExtProcessorTable {
...
+            st5_neq_st6() * (curr_row(ST6) - curr_row(ST5)),
+            st5_neq_st6() * (next_row(IP) - curr_row(JSO)),
+            st5_neq_st6() * (next_row(JSP) - curr_row(JSP) + one()),
+        ];
+        let maybe_recurse = vec![
+            st5_eq_st6() * (next_row(IP) - curr_row(JSD)),
+            st5_eq_st6() * (next_row(JSP) - curr_row(JSP)),
+            st5_eq_st6() * (next_row(JSO) - curr_row(JSO)),
+            st5_eq_st6() * (next_row(JSD) - curr_row(JSD)),
+        ];
```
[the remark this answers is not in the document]
jan-ferdinand: Keen observation! :ok_hand:

jan-ferdinand:
When targeting a degree of 12, which might be possible through a future upgrade, the additional cost of instruction `recurse_or_return` _with_ immediate argument becomes negligible: it introduces only 2 additional degree-lowering columns.

image

Instruction `recurse_or_return` _without_ immediate argument introduces XX additional degree-lowering columns:

image

Should the target degree be 12 at some point in the future, re-introducing the instruction's immediate argument could be considered.

PR #295, opened 2024-05-28
Halve combination codeword checks

### LINKED ISSUE #285: Separate Batch-Verification from First FRI Fold
A discussion with @ulrich-haboeck highlights a discrepancy between TritonVM's workflow and the workflow of (to wit) every other STARK construction. Other STARK systems separate the batch verification from the first folding step of FRI. In Triton VM, these steps are merged into one, and the first FRI codeword is never explicitly committed to. In exchange for introducing a new Merkle tree and the workloads of having to build and open it, and verify its paths, this alternative STARK construction reduces by half the number of rows of the low-degree-extended trace that must be opened and verified. This affects both the zero-knowledge calculus (fewer zerofiers required) and the verifier work (fewer inner products).

Sword-Smith:
- Closes #285 by halving the number of combination codeword checks performed outside of FRI.
- Fixes the number of trace randomizers to take revealed quotient segments into account

### DIFF
--- a/triton-vm/src/fri.rs
+++ b/triton-vm/src/fri.rs
@@ -128,12 +128,6 @@ impl<'stream, H: AlgebraicHasher> FriProver<'stream, H> {
             .sample_indices(indices_upper_bound, self.num_collinearity_checks);
     }
 
-    fn all_top_level_collinearity_check_indices(&self) -> Vec<usize> {
-        let a_indices = self.first_round_collinearity_check_indices.clone();
-        let b_indices = self.collinearity_check_b_indices_for_round(0);
-        a_indices.into_iter().chain(b_indices).collect()
-    }
-
     fn collinearity_check_b_indices_for_round(&self, round_number: usize) -> Vec<usize> {
         let domain_length = self.rounds[round_number].domain.length;
         self.first_round_collinearity_check_indices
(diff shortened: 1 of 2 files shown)


triton-vm/src/fri.rs
```diff
@@ -562,8 +546,7 @@ impl<H: AlgebraicHasher> Fri<H> {
         // it is important to modify the sponge state the same way.
         prover.proof_stream.sample_scalars(1);
 
-        let indices = prover.all_top_level_collinearity_check_indices();
-        Ok(indices)
+        Ok(prover.first_round_collinearity_check_indices.clone())
```
jan-ferdinand, line 549:
```suggestion
        Ok(prover.first_round_collinearity_check_indices)
```

nit: `prover` is consumed here, making the `.clone()` superfluous.

triton-vm/src/fri.rs
```diff
@@ -624,7 +607,8 @@ impl<H: AlgebraicHasher> Fri<H> {
         let num_rounds_checking_most_locations = num_rounds_checking_all_locations + 1;
 
         let num_rounds = max_num_rounds.saturating_sub(num_rounds_checking_most_locations);
-        num_rounds.try_into().unwrap()
+
+        usize::try_from(num_rounds).unwrap()
```
jan-ferdinand, line 611: nit: With the `.max(1)` gone again, this doesn't need changing.

jan-ferdinand (approved):
Nice. :clap: I left two nits in-line. I'm happy to apply them, then rebase & merge myself if you want.

PR #296, opened 2024-05-29
Start VM with initialized Sponge state

jan-ferdinand:
Previously, it was necessary to initialize Triton VM's Sponge state manually before any other Sponge instruction could be used. Now, instructions `sponge_absorb`, `sponge_absorb_mem`, and `sponge_squeeze` can be executed without first executing instruction `sponge_init`.

This is a breaking change because the `VMState`'s field `sponge` is both public and changes type from `Option<Tip5>` to `Tip5`.

Specific reasons I have flagged you for review:
- @Sword-Smith: how much work would this cause downstream?
- @aszepieniec: do the updated constraints look sane?

Apart from these specific questions, I'm happy to receive general ∧/∨ other feedback, of course.

### DIFF
--- a/specification/src/hash-table.md
+++ b/specification/src/hash-table.md
@@ -119,17 +119,6 @@ Both types of challenges are X-field elements, _i.e._, elements of $\mathbb{F}_{
 1. For `i` $\in \{0, \dots, 3\}$ and `limb` $\in \{$`highest`, `mid_high`, `mid_low`, `lowest` $\}$:<br />
     `state_i_limb_LookupClientLogDerivative` has accumulated `state_i_limb_lkin` and `state_i_limb_lkout` with respect to challenges 🍒, 🍓 and indeterminate 🧺.
 
-### Initial Constraints as Polynomials
-
-1. `Mode - 1`
-1. `round_no`
-1. `RunningEvaluationReceiveChunk - 🪣 - (🪑^10 + state_0·🪑^9 + state_1·🪑^8 + state_2·🪑^7 + state_3·🪑^6 + state_4·🪑^5 + state_5·🪑^4 + state_6·🪑^3 + state_7·🪑^2 + state_8·🪑 + state_9)`
-1. `RunningEvaluationHashInput - 1`
-1. `RunningEvaluationHashDigest - 1`
-1. `RunningEvaluationSponge - 1`
-1. For `i` $\in \{0, \dots, 3\}$ and `limb` $\in \{$`highest`, `mid_high`, `mid_low`, `lowest` $\}$:<br />
-    `state_i_limb_LookupClientLogDerivative·(🧺 - 🍒·state_i_limb_lkin - 🍓·state_i_limb_lkout) - 1`
-
 ## Consistency Constraints
 
 1. The `Mode` is a valid mode, _i.e._, $\in \{0, \dots, 3\}$.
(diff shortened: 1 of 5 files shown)


triton-vm/src/table/hash_table.rs
```diff
@@ -765,36 +781,43 @@ impl ExtHashTable {
...
+
+        // If the next round number is 0, then
+        // - the next mode is either `Hash` or `Pad`, or
+        // - the next instruction is `sponge_init`, or
+        // - the capacity does not change, or
+        // - the current mode is `ProgramHashing` and (!) the next mode is `sponge`.
+        //
+        // The “and” marked with an exclamation mark requires above constraint to be
+        // expressed through two polynomials.
+        let capacity_doesnt_change_at_section_start_when_program_hashing_or_ =
```
aszepieniec, line 800: I think `section_start` is a poor choice of words. What is a section? Are there at most four sections, one for each mode? I gather (perhaps incorrectly) that `section` means the same as `round`. Why not use that word instead? Or even better: `new_round`.

aszepieniec: If it does mean "contiguous region of constant mode" I would suggest using instead `mode_switch` instead.

aszepieniec: This (pair of) constraints is very confusing. Is it because it is trying to achieve several features at once? If so, consider separating it into more constraints. If mode switches from A to B then enforce C; if mode switches from X to Y then enforce Z; etc.

triton-vm/src/table/hash_table.rs
```diff
@@ -765,36 +781,43 @@ impl ExtHashTable {
...
-            .sum();
-        let if_round_number_next_is_0_and_ci_next_is_squeeze_then_state_doesnt_change =
+        let capacity_doesnt_change_at_section_start_when_program_hashing_or_absorbing_0 =
+            capacity_doesnt_change_at_section_start_when_program_hashing_or_.clone()
+                * Self::select_mode(circuit_builder, &mode, HashTableMode::ProgramHashing);
+        let capacity_doesnt_change_at_section_start_when_program_hashing_or_absorbing_1 =
+            capacity_doesnt_change_at_section_start_when_program_hashing_or_
+                * Self::select_mode(circuit_builder, &mode_next, HashTableMode::Sponge);
+
+        let randomized_sum_of_state_differences =
```
aszepieniec, line 814: Consider renaming to `state_doesnt_change`.

triton-vm/src/table/hash_table.rs
```diff
@@ -765,36 +781,43 @@ impl ExtHashTable {
...
+            izip!(&state_weights, &state_current, &state_next)
+                .map(|(weight, curr, next)| weight.clone() * (next.clone() - curr.clone()))
+                .sum();
+        let if_next_round_is_0_and_ci_next_is_squeeze_in_sponge_section_then_state_doesnt_change =
             Self::round_number_deselector(circuit_builder, &round_number_next, 0)
+                * Self::select_mode(circuit_builder, &mode, HashTableMode::ProgramHashing)
```
aszepieniec, line 820: I would expect there to be a *deselector* for mode `Sponge`. Can't you transition from something other than `ProgramHashing` to `Sponge`?

aszepieniec: > The four separate modes are `program_hashing`, `sponge`, `hash`, and `pad`, and they evolve in that order.

Got it!

specification/src/hash-table.md
```diff
@@ -189,8 +157,12 @@ define `state_i_hi_limbs_minus_2_pow_32 := 2^32 - 1 - 2^16·state_i_highest_lk_i
 1. If the `round_no` in the next row is 0
 and the `Mode` in the next row is either `program_hashing` or `sponge`
 and the instruction in the next row is either `sponge_absorb` or `sponge_init`,
-then the capacity's state registers don't change.
-1. If the `round_no` in the next row is 0 and the current instruction in the next row is `sponge_squeeze`, then none of the state registers change.
+then the capacity's state registers don't change,
```
aszepieniec, line 160: The capacity elements do not change if the next instruction is `sponge_init`? That doesn't sound right. I would expect them to be set to zero.

specification/src/hash-table.md
```diff
@@ -189,8 +157,12 @@ define `state_i_hi_limbs_minus_2_pow_32 := 2^32 - 1 - 2^16·state_i_highest_lk_i
...
+or the `Mode` in the current row is `program_hashing` and the `Mode` in the next row is `sponge`.
+1. If the `round_no` in the next row is 0
+and the current instruction in the next row is `sponge_squeeze`,
```
aszepieniec, line 163: "current instruction in the next row" is confusing; consider abbreviating to "CI".

specification/src/hash-table.md
```diff
@@ -189,8 +157,12 @@ define `state_i_hi_limbs_minus_2_pow_32 := 2^32 - 1 - 2^16·state_i_highest_lk_i
...
+then none of the state registers change,
+or the `Mode` in the current row is `program_hashing`.
```
aszepieniec, line 165: So if the current mode is `program_hashing` and the next instruction is `sponge_squeeze` then the state registers can change? Very counter-intuitive.

Issue #259, opened 2024-03-26
Instruction Maybe-Wishlist for Recursion and Consensus


aszepieniec:
This is a tracking issue. We add imagined instructions that could make recursion (or consensus programs) faster.

## `read_mem_forward`

 - BEFORE: `*ptr`
 - AFTER: `(*ptr+n) [element]` 
 - With this instruction, 7 instructions inside the inner loop of (a static variant of) `HashVarlen` becomes 3.

## `dot_step`

 - BEFORE: `_ acc2 acc1 acc0 *lhs *rhs`
 - AFTER: `_ acc2' acc1' acc0' (*lhs+3) (*rhs+3)`
 - With this instruction, 8 instructions inside the inner loop of `InnerProductOfThreeRowsWithWeights` becomes 1. Stands to reduce `1M to ~65000 cycles.

## `get_colinear_y`

 - BEFORE: `_ ax [ay] bx [by] [cx]` (possibly different order)
 - AFTER: `_ [cy]`
 - Where $c_y = \frac{a_y-b_y}{a_x-b_x} \cdot (c_x - a_x) + a_y$
 - This instruction can reduce the inner loop of `compute_c_values` from 74 instructions to 49. Total cycle count reduction: ~2000.

Sword-Smith:
It turns out the the opstack table height is the bottle-neck in the verifier, not the clock cycle count (which is recorded by the processor table). So counting number of clock cycles that a new instruction would shave off might not be the relevant metric.

I was considering a `recurse` instruction that acted as the current `recurse` if the top to stack elements are different and acts as `return` if the two top stack elements are the same. That would probably shave off a good amount of both opstack table rows since the start of almost all our loops looks like this:

```tasm
loop_label:
    dup m
    dup n
    eq
    skiz return
    
    loop body...
```

Sword-Smith:
The recurse_or_return I had thought of was one that returned if ST[5] == 0. Otherwise it recurses. That could be used in combination with merkle_step to reduce the loop body to 2 instructions.

aszepieniec:
I am observing that the following pattern occurs quite often:
```
                push {DIGEST_LENGTH - 1}
                add
                read_mem {DIGEST_LENGTH}
                pop 1
```

Maybe this can be shrunk to one instruction, `load_digest`?

aszepieniec:
Also the following:
```
                dup 8
                push 1 add
```
[...]

PR #297, opened 2024-05-30
Introduce instructions `pick` + `i` and `place` + `i`

jan-ferdinand:
The two new instructions simplify manipulation of the operational stack. Instruction `pick` + `i` moves the indicated stack element to the top of the stack. Instruction `place` + `i` is its dual, moving the top of the stack to the indicated position.

|                 | old op stack        | new op stack          |
|:----------------|:--------------------|:----------------------|
| `pick`  + `i`   | e.g., `_ d x c b a` | e.g., `_ d c b a x`   |
| `place`  + `i`  | e.g., `_ d c b a x` | e.g., `_ d x c b a`   |


jan-ferdinand:
Arguments in favor of introducing the new instructions:

- The assembly code of the recursive verifier in `tasm-lib` currently has 570 occurences of the pattern `swap 1 swap 2 swap 3`, which corresponds to `pick 3`, and 566 occurences of the pattern `swap 3 swap 2 swap 1`, which corresponds to `place 3`.[^1]
- A common pattern for reading from memory when the memory pointer is in the shallow part of the stack is `swap i read_mem swap j pop 1` which can be replaced by `pick i read_mem place j`.
- “swap juggling” is an acquired skill and might be a deterrent for newcomers.

Arguments against introducing the new instructions:

- Arithmetic overhead.

[^1]: This is _not_ the number of times these patterns are executed, _only_ how often they appear in the source code. The number of times these patterns are executed might be much higher and requires further analysis.

aszepieniec:
To make the strongest possible argument in favor of this PR, one has to argue that the following cost metrics either don't increase or are worth it:
 - the number of new constraints (recall that we need to compute a linear combination with this number of terms; so it affects prover speed)
 - the number of new degree-lowering columns
 - the reduced space of opcodes.

Do we have numbers or arguments relating to these cost metrics?

jan-ferdinand:
The arithmetization overview can answer the first two questions:
- number of new constraints: 4
- number of new degree-lowering columns: 4

The arithmetization overview also tells us that evaluating the new constraints would introduce 2382 additional rows in the processor table when evealuating the AIR in assembly.

The question regarding opcode space is slightly more involved because of [instruction categories](https://triton-vm.org/spec/instructions.html#regarding-opcodes). The relevant categories are currently (without `pick` and `place`):

| u32?    | shrink?      | size        | num |
|:--------|:-------------|:------------|----:|
| non-u32 | no shrink    | single word |  13 |
| non-u32 | no shrink    | double word |   7 |
| non-u32 | shrink stack | single word |  11 |
| non-u32 | shrink stack | double word |   3 |
| u32     | no shrink    | single word |   4 |
| u32     | no shrink    | double word |   0 |
| u32     | shrink stack | single word |   4 |
| u32     | shrink stack | double word |   0 |

Within each category, the maximum is 16. The instruction `pick` and `place` both fall in the category (non-u32, no-shrink, double word), increasing its size from 7 to 9.

aszepieniec (approved):
Reviewed. No comments. LGTM.

aszepieniec:
To merge or not to merge? Here is an argument in favor worth considering.

The 1.6% fewer cycles translates to rows in the processor table. Assume that translates to a smaller table height before padding. (And even if this assumption is wrong the argument still kinda holds.) We get this reduced row count in exchange for 0.07% (=1 - 607/611) performance drop in two big steps, namely NTT and row hashing. While this PR does not move the row count across the next power-of-two threshold, it can achieve that *in combination with other similar improvements*.

I propose to merge this and future similar improvements. And then, at a later stage when we crossed the next power of two boundary, we can figure out which instructions to drop without crossing it back. In the mean time we can
 - avoid using the instructions explicitly so as to not complicate the potential reversion
 - automatically optimize generated tasm so as to reap the benefits.

Sword-Smith:
> To merge or not to merge? Here is an argument in favor worth considering.
> 
> The 1.6% fewer cycles translates to rows in the processor table. Assume that translates to a smaller table height before padding. (And even if this assumption is wrong the argument still kinda holds.) We get this reduced row count in exchange for 0.7% (=1 - 607/611) performance drop in two big steps, namely NTT and row hashing. While this PR does not move the row count across the next power-of-two threshold, it can achieve that _in combination with other similar improvements_.

How did @jan-ferdinand  get to the 1.6 % fewer cycles? Do they include the added cost of the AIR-constraint evaluation? Because of this AIR-constraint evaluation cost, this PR adds 2 % to the opstack table 0.8 % to the processor table, and 0.6 % to the memory table. With the above-mentioned 0.7 % of additional prover time, I would say merging this is neutral from a performance perspective (assuming I didn't misunderstand anything).

I'm in favor of advancing on the consensus programs and only after those have been written, reconsider these instruction. But it's not a strong preference. I'm happy to let the majority decide.

Sword-Smith:
For reference, here is a TVM-benchmark result for the current verifier, run on Triton-VM 0.42 alpha 6.

```json
[
  {
    "name": "tasmlib_verifier_stark_verify_inner_padded_height_524288_fri_exp_4",
    "benchmark_result": {
      "clock_cycle_count": 285183,
      "hash_table_height": 234025,
      "u32_table_height": 211197,
      "op_stack_table_height": 235124,
      "ram_table_height": 281599
    },
    "case": "CommonCase"
  }
```

jan-ferdinand:
Please note that [yesterday's comment](https://github.com/TritonVM/triton-vm/pull/297#issuecomment-2247290419) was not meant to try to conclude the question regarding merging this PR; I mostly wanted to jot down some findings wherever they fit best. Some of this is here.
However, more of the findings are in [this comment](https://github.com/TritonVM/triton-vm/issues/259#issuecomment-2248270115) on #259, where I try to analyse the recufier's execution trace in order to identify common patterns that might be worth introducing additional instructions for. The method is [somewhat crude](https://github.com/TritonVM/tasm-lib/blob/analysis/tasm-lib/count_patterns.py) right now, and might not be the right tool for all questions in this line.

jan-ferdinand:
> How did @jan-ferdinand get to the 1.6 % fewer cycles?

I analysed how often a `swap` pattern that could be replaced by a `pick` & `place` pattern occured when executing the recufier, then used the respective patterns lengths as well as the total number of executed instructions.

> Do the [1.6% reduced number of instructions] include the added cost of the AIR-constraint evaluation?

No. That's part of the (admittedly overly short) con “arithmetic overhead.”

PR #299, opened 2024-06-08
Add prove verify command

### LINKED ISSUE #238: add command line tool for proof generation & verification
It would be nice to run Triton VM as a command-line tool; Andrew Milson's [demo of ministark](https://github.com/andrewmilson/ministark#demo---proving-brainfk-programs) is a good example; basically one sub-command for proving and dumping something into a file (or stdout), and another sub-command for verifying by reading that file (or stdin).

_(extracted from #164)_

cyberbono3:
Resolves #238

```
cargo run --example cmd prove --dst ./triton-vm/examples/proof.txt
Proof generated in: 43s
Proof size: 1820KB
Stark, claim and proof written to ./triton-vm/examples/proof.txt
```

```
cargo run --example cmd verify ./triton-vm/examples/proof.txt     
stark length: 103
claim length: 148
Proof verified in: 175.564791ms
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ trybuild = "1.0"
 twenty-first = "0.42.0-alpha.3"
 unicode-width = "0.1"
 
+
 [workspace.lints.clippy]
 cast_lossless = "warn"
 cloned_instead_of_copied = "warn"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ twenty-first = "0.42.0-alpha.3"
 unicode-width = "0.1"
 
 
+
 [workspace.lints.clippy]
 cast_lossless = "warn"
 cloned_instead_of_copied = "warn"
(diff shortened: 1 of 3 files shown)


cyberbono3:
> Thanks for the PR!
> 
> I think this is a decent example showcasing how proofs can be stored on disk, read from disk, and subsequently verified. For a full resolution of #238, I think the main shortcomings are:
> 
> * The program to be proven & verified is hardcoded. For a full-fledged CLI, the program (as well as input and non-determinism) should be given by the user at runtime.
> * Being an example, invocation of the CLI is a bit awkward. To me, it feels more natural to write `triton-cli prove my_program.tasm` than `cargo run --example cmd prove my_program.tasm`. If the PR added a crate (for example in the same workspace), that crate could be a standalone binary. This binary could then even be installed.
> 
> An additional note: As `structopt` is not actively developed anymore and [superseded](https://docs.rs/structopt/latest/structopt/#maintenance) by [`clap`](https://crates.io/crates/clap), I think it is preferrable to use `clap` as the command line argument parser.
> 
> Should you have questions regarding some of these steps, feel free to ask. Also, please don't feel obligated by any of this; adding an example as a starting point for a full-fledged CLI is fine addition by itself.
> 
> Would you want to keep working on this?

yes!

PR #305, opened 2024-07-11
feat: triton-cli prove/verify, proof binary serialization

### LINKED ISSUE #238: add command line tool for proof generation & verification

chancehudson:
This PR adds a new crate `triton-cli`. This crate exposes a cli with two commands
- `prove`: read asm from file, prove it, and write the resulting proof to file
- `verify`: read a proof from file, verify it, and print details about it

Proofs are serialized in a simple binary format. This format is described in the `main.rs` for the cli crate.

From inside the `triton-cli` directory run `cargo install --path .` to add the `triton-cli` to your path.

```zsh
> triton-cli prove ./simple.asm out.proof --private-inputs=12412414,9 --public-inputs 214
proving...
success!
proof written to: out.proof
>
```

```zsh
> triton-cli verify out.proof
proof is valid!
program digest: 0xbd500d8c2bc5b683fe957fa7e2525808b1eb113ad21a99044eec82abdff548e6f27d8b1429c6eaef
=================
security level: 160 bits
FRI expansion factor: 4
trace randomizers: 166
colinearity checks: 80
codeword checks: 160
=================
public inputs:
(none)
public outputs:
(none)
>
```

Things to watch out for:
- proof data is written non-atomically, halting the process can create a corrupt proof file
- inputs are not implemented
- the binary format describes lengths as "number of 64-bit words" instead of number of bytes

Closes #238


vimwitch:
Hey @aszepieniec thanks for the review. I based the cli on the feedback in #299. I wanted to see if there were any ideas on how to implement inputs (private and public). Naively i'd probably implement it as two optional flags on the prove command, each accepting a comma separated list of values. The next step from that would be passing files in some format.

vimwitch:
This PR isn't time sensitive though, i've got a local copy of the cli i'm using for debugging so there's no rush here

triton-cli/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+    // proof_length: u64
+    //
+    // program_digest: u64[digest_length]
+    //
+    // public_inputs: u64[public_input_length]
+    // public_outputs: u64[public_output_length]
+    // proof: u64[proof_length]
+    write_proof(data, out).expect("failed to write proof to file");
+    println!("proof written to: {out}");
+}
```
jan-ferdinand, line 136: Instead of (de)serializing ad-hoc, how about using libraries built for exactly that purpose instead? In the example below, I'm using `serde` and `bincode`, but other alternatives might exist.

```rust
#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
struct SerializableProof {
    version: u8,
    stark: Stark,
    claim: Claim,
    proof: Proof,
}

fn write_proof(proof_path: &str, (stark, claim, proof): (Stark, Claim, Proof)) -> Result<()> {
    let serializable_proof = SerializableProof {
        version: 1,
        stark,
        claim,
        proof,
    };
    let encoded = bincode::serialize(&serializable_proof)?;
    fs::write(proof_path, &encoded)?;
    Ok(())
}

fn read_proof(proof_path: &str) -> Result<(Stark, Claim, Proof)> {
    let encoded = fs::read(proof_path)?;
    let SerializableProof {
        version,
        stark,
        claim,
        proof,
    } = bincode::deserialize(&encoded)?;
    if version != 1 {
        bail!("unsupported proof version: {version}"); // uses crate `anyhow`
    }
    Ok((stark, claim, proof))
}
```

This is probably not the final version of the code yet: in the suggestion, the proof's version is embedded in the struct `SerializableProof`. However, in order to actually support different proof versions, it should probably be parsed first, and subsequent parsing should then depend on the version.

triton-cli/test-vectors/simple.tasm
```diff
@@ -0,0 +1,1740 @@
+push 10
```
jan-ferdinand, line 1: A file with 1740 lines doesn't exactly scream “simple” to me… :astonished: Would it make sense to use something smaller for testing?

triton-cli/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+        .collect::<Vec<BFieldElement>>()
+}
+
+fn prove(
+    asm: &String,
+    out: &String,
+    public_inputs: Option<String>,
+    private_inputs: Option<String>,
+) {
+    // TODO: test paths before making the proof?
```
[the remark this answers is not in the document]
jan-ferdinand: Also, silently overwriting an existing file might not make everybody happy. It could be a possibility to abort & warn the user if the output file already exists. I'm not sure where I stand on this, though; other CLI utilities silently overwrite existing files also, so there certainly is precedence.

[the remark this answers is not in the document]
vimwitch: It's the opposite, if the file exists it throws an error. But this is almost worse because if a large proof is built it will be discarded if the output filepath already exists.

triton-cli/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+    println!("trace randomizers: {}", stark.num_trace_randomizers);
+    println!("colinearity checks: {}", stark.num_collinearity_checks);
+    println!("codeword checks: {}", stark.num_combination_codeword_checks);
+    println!("=================");
+    println!("public inputs:");
+    if claim.input.is_empty() {
+        println!("(none)");
+    }
+    for v in claim.input {
+        println!("{}", u64::from(v));
```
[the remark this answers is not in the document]
vimwitch: Nope, I'm a rust noob and i've never read the `Display` docs and did not know it was implemented for this type 😎

triton-cli/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+fn digest_to_str(d: Digest) -> String {
+    let mut hex = String::new();
+    d.0.iter()
+        .for_each(|v| write!(&mut hex, "{:16x}", u64::from(v)).unwrap());
+    format!("0x{hex}")
+}
+
+fn verify(proof_path: &String) {
```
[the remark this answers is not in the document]
vimwitch: Great tip, thanks!

triton-cli/test-vectors/simple.tasm
vimwitch: Yea it's a short program that i duplicated with a vim macro. I removed all but 25 lines.

triton-cli/src/main.rs
vimwitch: > This is probably not the final version of the code yet: in the suggestion, the proof's version is embedded in the struct SerializableProof. However, in order to actually support different proof versions, it should probably be parsed first, and subsequent parsing should then depend on the version.

Yes, i really just wanted to reserve the first byte so in the future it's possible to implement versioning.

I didn't use a library because i'm not familiar with them. In the example above, will changes to the `Stark`, `Claim`, or `Proof` struct (or any nested structs) change the file structure? I'm not opposed to using libraries but I think it's important that the file structure is well documented/stable so others can write parsers in other languages. Explicitly writing the file byte by byte is an easy way to do that.

vimwitch: I guess my question is, how do we make sure `version` is changed anytime the `Stark`/`Claim`/`Proof` struct is modified?

triton-cli/src/main.rs
```diff
@@ -0,0 +1,259 @@
...
+    for v in claim.output {
+        println!("{}", u64::from(v));
+    }
+}
+
+fn parse_inputs(inputs: Option<String>) -> Vec<BFieldElement> {
+    inputs
+        .unwrap_or_default()
+        .split(',')
+        .filter(|v| !v.is_empty())
```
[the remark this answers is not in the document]
vimwitch: If `inputs` is `None` `unwrap_or_default` returns `""`, which after being run though `split` ends up as `[""]`.

I didn't think it was a problem to allow e.g. `--public-inputs=,`, and it avoided adding a conditional.

vimwitch:
@jan-ferdinand great review, thank you

> initial RAM, which can be used by any instruction accessing memory

I was just about to open an issue asking about initial RAM. I'm not sure how to specify this for a program, do you have any links that could help me?

> non-deterministic digests, which can be accessed by instruction divine_sibling

Also something I'm not familiar with. I've read the `divine_sibling` docs but I didn't fully understand how the instruction is meant to be used.

I'm kinda thinking that these additional methods of specifying inputs should be added in separate pr's.

> The user experience might become better if the error-propagation crate anyhow was used. By changing signature of main() to return an anyhow::Result<()>, errors can be propagated all the way to the program's entrypoint. This removes the need for many a .unwrap() or .expect().

Looking into this rn

vimwitch:
The `anyhow` crate is great. I've been struggling to figure out clean error handling and this crate basically solves it

triton-cli/src/main.rs
vimwitch: I ended up making a custom struct and using serde+bincode with that.

triton-cli/src/main.rs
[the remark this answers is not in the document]
jan-ferdinand: In that case, see if you like what `Display` for `BFieldElement` does; one “quirk” is that field elements 18446744069414584066 through 18446744069414584320 are printed as -255 through -1, which can be handy but can also be confusing. :relieved: Mathematically, it's correct, but that's not all that matters in this case. :slightly_smiling_face:

triton-cli/src/main.rs
jan-ferdinand: > I guess my question is, how do we make sure `version` is changed anytime the `Stark`/`Claim`/`Proof` struct is modified?

Yeah, that's a super valid question. My proposed “solution” falls short on at least two counts:
- since the `version` is a field on the struct itself, making de-serialization conditional to the version doesn't work
- it's not obvious (to me) how to ensure correct versioning if any of `Stark`, `Claim`, or `Proof`'s format changes

A custom (de)serializer certainly allows addressing both of those shortcomings. That said, it feels as if this is a problem other people have encountered before, and have provided a general solution for. Maybe this solution is `serde` ∧/∨ `bincode`, but I'm not sure. I'll take another look. :relieved:

jan-ferdinand: > Yes, i really just wanted to reserve the first byte so in the future it's possible to implement versioning.

That is a super good idea as far as I can see. :+1:

triton-cli/src/main.rs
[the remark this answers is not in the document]
jan-ferdinand: I see. So maybe that line is catching more than you had intended? It seems that the current `.filter()` makes `--public-input=5,,,,10` valid. Do you want it to be?

jan-ferdinand:
> The `anyhow` crate is great. I've been struggling to figure out clean error handling and this crate basically solves it

Should you ever be looking for something similar but for a library, check out [`thiserror`](https://crates.io/crates/thiserror). Same author, and also super convenient to use. Both `thiserror` and `anyhow` refer to each other in their readmes, so it's easy to look up which one is better suited for which use case. :relieved:

jan-ferdinand:
> @jan-ferdinand great review, thank you

Happy to! :blush:


> I'm not sure how to specify [initial RAM] for a program, do you have any links that could help me?

Yeah, the docs on this are not super great, I have to admit. :bow: There are various answers to your question.
- `NonDeterminism` has a method `fn with_ram(self, ram: HashMap<BFieldElement, BFieldElement>) -> Self`[^1] which allows to to specify the initial RAM.
- It's not obvious how to let a user specify initial RAM as conveniently as possible. [Triton TUI](https://github.com/TritonVM/triton-tui/) uses a JSON file for all three types of private input, [for example like so](https://github.com/TritonVM/triton-tui/blob/master/examples/non_determinism.json). When starting the TUI, the path to such a JSON file can be given as [one of the command line arguments](https://github.com/TritonVM/triton-tui/blob/master/src/args.rs#L64). There might be better ways to go about this and I'm happy to discuss alternatives. :relieved:

> I've read the `divine_sibling` docs but I didn't fully understand how the instruction is meant to be used.

Yeah, `divine_sibling` is a bit of a special kid. The intention is to simplify verification of inclusion of elements in a [Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree). Without the context of Merkle trees, the instruction probably won't make a whole lot of sense. That said, let's take the following Merkle tree:

image

The node's indices are green, the parent/child relation is marked by the black lines. The tree's root has index 1.[^2] Let's assume you want to walk the authentication path for node 10. In the picture, the authentication path in question is marked by red circles. To compute node 10's parent, which is node 5, you also need to know node 11. Hashing left sibling (node 10) and right sibling (node 11) gives the parent (node 5). In order to get node 5's parent, you also need node 4. Again, hashing left sibling (node 4) and right sibling (node 5), gives the parent (node 2).
Crucially, it is important that the left sibling is on the left; hashing right sibling and left sibling (concatenated in this order) gives a different result from hashing left sibling and right sibling (in this order).

Now, instruction `divine_sibling` takes care of a few things in order to simplify walking an authentication path. Note how sometime, the already-known-sibling is on the left, and sometimes, it is on the right.[^3] Instruction `divine_sibling` takes care that the known sibling is moved to the correct side, and that the unkown sibling is “divinely” guessed and put to the other side. Then, calling instruction `hash` gives the parent node. That means that moving up three layers of a Merkle tree is as easy as writing `divine_sibling hash divine_sibling hash divine_sibling hash`.[^4]

Of course, this produces the correct result only if the private input for sibling digests, also known as “non-deterministic digests”, is correctly populated. Feel free to ask additional questions. :relieved:

> I'm kinda thinking that these additional methods of specifying inputs should be added in separate pr's.

That's fine with me. :slightly_smiling_face::+1: 

[^1]: the signature is slightly more general, but you get the idea
[^2]: For Merkle trees, 1-indexing is easier than 0-indexing, hence a break from the norm.
[^3]: The least significant bit of the node index is a fast way to identify which side the sibling is on.
[^4]: Upcoming instruction `merkle_step` simplifies this to `merkle_step merkle_step merkle_step`.

triton-cli/src/main.rs
```diff
@@ -89,171 +89,122 @@ fn parse_inputs(inputs: Option<String>) -> Vec<BFieldElement> {
...
 
-fn read_proof(proof_path: &String) -> Result<(Stark, Claim, Proof)> {
-    let file = fs::File::open(proof_path)?;
-    let mut reader = BufReader::new(file);
-    let version = read_u8(&mut reader);
-    assert!(version == 1, "wrong proof file version!");
+fn read_proof(proof_path: &str) -> Result<(Stark, Claim, Proof)> {
+    let mut file = fs::File::open(proof_path)?;
+    let mut proof_bytes: Vec<u8> = Vec::new();
+    file.read_to_end(&mut proof_bytes)?;
```
jan-ferdinand, line 139: From `fs::read`:

```md
/// Read the entire contents of a file into a bytes vector.
///
/// This is a convenience function for using [`File::open`] and [`read_to_end`]
/// with fewer imports and without an intermediate variable.
```

```suggestion
	let proof_bytes = fs::read(proof_path)?;
```

triton-cli/src/main.rs
```diff
@@ -89,171 +89,122 @@ fn parse_inputs(inputs: Option<String>) -> Vec<BFieldElement> {
...
-    let digest_len = read_u8(&mut reader);
-    let input_len = read_u64(&mut reader);
-    let output_len = read_u64(&mut reader);
-    let proof_len = read_u64(&mut reader);
-    let mut d = Digest::default();
+    let mut digest = Digest::default();
     assert_eq!(
-        d.0.len(),
-        usize::from(digest_len),
+        digest.0.len(),
```
jan-ferdinand, line 152: Note that there is a `const DIGEST_LENGTH` (soon an associated `const Digest::LEN`), which seems more suitable.

jan-ferdinand: Even easier is probably to use one `Digest`s `try_from` methods directly, which will return an error if the length does not match.

```rust
let digest = Digest::try_from(&program_digest)?;
```

triton-cli/src/main.rs
```diff
@@ -89,171 +89,122 @@ fn parse_inputs(inputs: Option<String>) -> Vec<BFieldElement> {
...
+    public_inputs: Vec<u64>,
+    public_outputs: Vec<u64>,
+    proof: Vec<u64>,
 }
 
-fn read_proof(proof_path: &String) -> Result<(Stark, Claim, Proof)> {
-    let file = fs::File::open(proof_path)?;
-    let mut reader = BufReader::new(file);
-    let version = read_u8(&mut reader);
-    assert!(version == 1, "wrong proof file version!");
```
jan-ferdinand: Was the deletion of this line accidental, or do you simply accept any version specificer for now, anticipating restrictions in the future?

triton-cli/src/main.rs
```diff
@@ -89,171 +89,122 @@ fn parse_inputs(inputs: Option<String>) -> Vec<BFieldElement> {
...
+fn write_proof(data: (Stark, Claim, Proof), out: &str) -> Result<()> {
     let (stark, claim, proof) = data;
+    let serialized = SerializedProof {
+        version: 1,
+        security_level: u64::try_from(stark.security_level)?,
+        fri_expansion_factor: u64::try_from(stark.fri_expansion_factor)?,
+        num_trace_randomizers: u64::try_from(stark.num_trace_randomizers)?,
+        num_colinearity_checks: u64::try_from(stark.num_collinearity_checks)?,
+        num_combination_codeword_checks: u64::try_from(stark.num_combination_codeword_checks)?,
+        program_digest: claim.program_digest.0.iter().map(u64::from).collect(),
```
jan-ferdinand, line 191: Maybe (but it's not important):

```suggestion
        program_digest: claim.program_digest.values().map(u64::from).collect(),
```

triton-cli/test-vectors/simple.tasm
```diff
@@ -0,0 +1,24 @@
...
+pop 1
+push 5
+swap 3
+pop 1
+push 5
+swap 3
+pop 1
+push 5
+swap 3
+halt
```
jan-ferdinand, line 24: nit: [POSIX says that a line must end with `\n`](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline). That will remove github's the scary-looking “🚫”. :smiley:

triton-cli/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+    Ok(())
+}
+
+#[test]
+fn test_serialization() -> Result<()> {
+    let asm = "./test-vectors/simple.tasm".to_string();
+    let proof = "./test-vectors/simple.proof".to_string();
+    prove(&asm, &proof, None, None)?;
+    verify(&proof);
+    fs::remove_file(proof).unwrap();
```
jan-ferdinand, line 208: Maybe:

```suggestion
    fs::remove_file(proof)?;
```

triton-cli/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+    file.write_all(&proof_bytes)?;
+    Ok(())
+}
+
+#[test]
+fn test_serialization() -> Result<()> {
+    let asm = "./test-vectors/simple.tasm".to_string();
+    let proof = "./test-vectors/simple.proof".to_string();
+    prove(&asm, &proof, None, None)?;
+    verify(&proof);
```
jan-ferdinand, line 207: Nit: you can leave the strings on the stack.

```suggestion
    let asm = "./test-vectors/simple.tasm";
    let proof = "./test-vectors/simple.proof";
    prove(asm, proof, None, None)?;
    verify(proof);
```

triton-cli/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+    let proof_vec = serialized_proof
+        .proof
+        .iter()
+        .map(|v| BFieldElement::new(*v))
+        .collect();
+    Ok((stark, claim, Proof(proof_vec)))
+}
+
+fn write_proof(data: (Stark, Claim, Proof), out: &str) -> Result<()> {
+    let (stark, claim, proof) = data;
```
jan-ferdinand, line 183: If you want, you can pattern match the input directly:

```suggestion
fn write_proof((stark, claim, proof): (Stark, Claim, Proof), out: &str) -> Result<()> {
```

It's a matter of taste, though, and any decision here is entirely yours. :relieved:

triton-cli/src/main.rs
```diff
@@ -89,171 +89,122 @@ fn parse_inputs(inputs: Option<String>) -> Vec<BFieldElement> {
...
-    // public_inputs: u64[public_input_length]
-    // public_outputs: u64[public_output_length]
-    // proof: u64[proof_length]
-    write_proof(data, out).expect("failed to write proof to file");
+    write_proof(data, out).with_context(|| "Failed to write proof to file")?;
     println!("proof written to: {out}");
+    Ok(())
+}
+
+#[derive(Serialize, Deserialize)]
```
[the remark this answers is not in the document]
vimwitch: It was, I didn't see a reason to derive them and thought there would be (marginal) overhead. Maybe it's a better habit to derive them by default/more aggressively? 🤔

triton-cli/src/main.rs
vimwitch: This `try_from` variant isn't working for me. The `program_digest` is a `Vec<u64>`. Even if i change it to a `Vec<u8>` it isn't compatible because `Digest` wants a `[u8; 40]`. I ended up mapping the `Vec<u64>` into `Vec<BFieldElement>` and pulling a slice from that, and it's slightly more semantic.

triton-cli/src/main.rs
vimwitch: accidental, good catch

triton-cli/src/main.rs
vimwitch: ah, didn't know i could do js style destructures 👍🏻

triton-cli/src/main.rs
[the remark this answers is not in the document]
vimwitch: Yes, i realized it made sequential commas valid. My thinking is, it allows two cases:
- `--public-inputs=11,22,33,,44,55` - the case where a user makes a typo
- `--public-inputs=11,22,33,` - the case where a user leaves a trailing comma (or leading)

This is ambiguous though, it's unclear if `--public-inputs=11,,22` results in `[11,22]` or `[11,0,22]` (the former is the actual parsed input).

I'm looking at is as a small quality of life thing even though it's not technically correct. I certainly don't mind changing it though.

vimwitch:
Great explanation of `divine_sibling`, thanks.

> Of course, this produces the correct result only if the private input for sibling digests, also known as “non-deterministic digests”, is correctly populated. Feel free to ask additional questions. 😌

The explanation makes sense, though i'll need to play with it more to understand it fully. My only question is, what motivated you to implement this at the VM level instead of the program level? Or rather, how much overhead (if any) does it add to programs that don't use the instruction? Or is it necessary for making the stark proof itself?

triton-cli/src/main.rs
jan-ferdinand: Ah, good point. I suppose we could add a `try_from` from `Vec<u64>` but I think the path you describe works well. :ok_hand:

triton-cli/src/main.rs
jan-ferdinand: I was happily surprised when I found out. :smiley: I mostly wanted to point out the possibility, please feel free to use either style. :relieved:

jan-ferdinand:
> what motivated you to implement this at the VM level instead of the program level?

It's an optimization for recursive verification and not required to generate a STARK proof. An alternative to having the instruction would be something along the lines of:

```asm
; never run, never tested – likely to contain bugs
divine 5	; get the sibling digest
dup 10		; duplicate the node index
push 2
div_mod
pop 1		; least-significant bit of node index is now on top of the stack
skiz
	call swap_digests ; requires something like 15 instruction? I didn't count
```

When verifying a STARK, there are many Merkle tree verifications happening, each of which requires several executions of `divine_sibling`. If we were to replace the instruction with the program outlined above, the blow-up factor of roughly 20 would hurt a lot. In fact, the main reason for introducing `merkle_step`[^1] and removing `divine_sibling` is how many instructions it saves us – and that's with a reduction of 2 $\rightarrow$ 1!

You are right in assuming that there is an overhead to having this instruction around. And unfortunately, this overhead is something that everyone using Triton VM is paying, independend of them using the instruction. It is possible to design ZK-VMs that mitigate this to a certain degree, but Triton VM has not adopted any of those mitigation strategies yet; our current focus is elsewhere. :relieved:
Quantifying the overhead is also not easy. It is possible to get numbers, but unfortunately, that requires significant changes to the code for the time being.

[^1]: `merkle_step` is equivalent to `divine_sibling` followed by `hash`

cyberbono3:
@chancehudson  Good job!

triton-cli/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+    security_level: u64,
+    fri_expansion_factor: u64,
+    num_trace_randomizers: u64,
+    num_colinearity_checks: u64,
+    num_combination_codeword_checks: u64,
+    program_digest: Vec<u64>,
+    public_inputs: Vec<u64>,
+    public_outputs: Vec<u64>,
+    proof: Vec<u64>,
+}
```
jan-ferdinand, line 133: Getting this right seems to be more difficult (for me) than I had anticipated. I see more or less three main ways we can go about it:

1. Declare proof versioning out of scope for Triton CLI.
2. Detect out-of-sync versions and report them as an error.
3. Support as many proof versions as possible.

## 1 – Declare Versioning to be Out of Scope

The simplest “solution” from an implementation standpoint. If deserialization fails, the cause might be an out-of-date proof, hard disk failure, aliens, …. The user can check for an update of Triton CLI and try their luck again.

## 2 – Error on Out-of-Sync Proof

Requires
- storing the proof's version alongside the proof, and
- knowing the version of Triton VM.

## 3 - Support Different Proof Versions

Requires everything from (2) plus old versions of Triton VM as additional dependencies. Then, proof generation / verification can be delegated to the chosen version. This is certainly possible by [renaming dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml). The biggest downside I see are compilation times and binary size – Triton VM is not a particularly leightweight library.

# Detecting Versions

## Of the Proof

For strategies 2 and 3, the proof's version must be accessible whether deserialization works or not. Here are suggestions that might work:
1. Use the file's first few bytes for the proof's version, store the serialized proof in `bytes[some_index..]`.
2. Declare a big `enum` with one struct variant per proof version. For example:
```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
enum VersionedSerializableProof {
	V0_13 { stark: triton_vm_0_13::Stark, claim:  triton_vm_0_13::Claim, .. },
	V0_22 { stark: triton_vm_0_22::Stark, claim:  triton_vm_0_22::Claim, .. },
	..
}
```
I'm not sure this works, though: I don't think `serde` is able to distinguish between two different versions of same type.

## Of Triton CLI / VM

With strategy (2) “Error on out-of-sync proof”, the version of Triton VM in use must be known. The ways I see to achieve this, are:
1. Use compile-time environment variable `CARGO_PKG_VERSION` via the `env!` macro to get a `&'static str` to the version specifier in the `Cargo.toml` of Triton CLI. Only works under the assumption that Triton CLI has the same version as Triton VM. Notably, this is not currently the case.
2. Use a `build.rs` build script and some parsing of the `Cargo.toml` to populate a compile-time environment variable with the version of Triton VM in use, which can then be read via the `env!` macro.
3. Manually update some hardcoded `const` whenever the version of Triton VM is bumped.

I must admit, none of these approaches seem particularly resilient.


What do you think?  Do these thoughts and conclusions make sense? Am I overlooking something blatantly obvious?

triton-cli/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        DIGEST_LENGTH,
+        serialized_proof.program_digest.len(),
+        "digest length mismatch!"
+    );
+    let digest_elements = serialized_proof
+        .program_digest
+        .iter()
+        .map(|v| BFieldElement::from(*v))
+        .collect::<Vec<BFieldElement>>();
+    let digest = Digest::try_from(&digest_elements[0..DIGEST_LENGTH])?;
```
jan-ferdinand, line 160: You don't have to do the length check yourself if you pass the entire `digest_elements` into the `try_from`:

```suggestion
    let digest_elements = serialized_proof
        .program_digest
        .iter()
        .map(|v| BFieldElement::from(*v))
        .collect::<Vec<_>>();
    let digest = Digest::try_from(digest_elements)?;
```

jan-ferdinand: In fact, you can remove a lot of `.iter().map(…)`ing by changing the types used in the `SerializedProof`:

```rust
#[derive(Serialize, Deserialize, Debug, Clone)]
struct SerializedProof {
	..
    program_digest: Digest,
    public_inputs: Vec<BFieldElement>,
    public_outputs: Vec<BFieldElement>,
    proof: Vec<BFieldElement>,
}
```

All of these types (de)serialize just fine. :slightly_smiling_face:

chancehudson: I applied the first change passing the full vector to `try_from`. I thought about changing the types in `SerializedProof`, but this makes the file structure vulnerable to change if `Digest` or `BFieldElement` changes. That's why i'm unwrapping/re-wrapping into primitive types.

triton-cli/src/main.rs
chancehudson: In versioning we have 2 things that can change. The first is the structure of the binary file itself. This affects how we implement encoding/decoding, but does not necessarily affect how the proof is verified. The second is the version of triton-vm that generated the proof. This is much more difficult to handle because a proof might be (de)serialized just fine and then fail verification because e.g. the merkle structure was off by 1 or the hash function changed etc.

The first case is firmly in scope for this package. Anytime we change the `SerializedProof` structure we should keep the old structure/implementation somewhere and `match` into the old structure based on the version indicator until we explicitly drop support for a version. The second case seems mostly out of scope for this package. By that i mean, we should not try and support anything but the current version. The triton-cli version number should follow triton-vm. e.g. anytime a breaking version of triton-vm is released, a breaking triton-cli version should be released as well. I would lean into the versioning system here and use the `CARGO_PKG_VERSION` to throw an error if a proof is verified from the wrong major version.

The tricky part here is _detecting_ when a breaking change is made to the STARK prover/verifier logic and strictly following versioning rules. I would commit a proof to this repo from the current version of triton-vm and have the CI attempt to verify this proof on every push. This will fail anytime a breaking change is made which is an indicator to push a breaking version release (or try to undo the breaking change). Sort of a makeshift e2e test.

I think to mitigate some of the risk here we can add more variables into the serialized proof. Namely:
- name/identifier of hash function used
- field/field extension used

This would add slightly more flexibility on detecting implementation details.

chancehudson: I would handle this versioning stuff lazily though. Let's call these things the "proof file version" and the "prover version". 

We start with proof file version 1, there are no other versions to support right now. We start with prover version `0.x.x`. We specify that if the prover version is not defined in the proof file it is assumed to be `0.x.x`.

To _pedantically_ support this I think we need to add the `CARGO_PKG_VERSION` check logic to the prover/verification and throw an error on prover version `>0.x.x`. It would be good to add a comment near this logic documenting this conversation/plan.

chancehudson: I applied this in [553a028](https://github.com/TritonVM/triton-vm/pull/305/commits/553a028b40c14e1a4c6aa7036adf9b7b8aa4c3d0).

I ended up putting the version in the proof file as it seemed more self-documenting.

chancehudson: BTW i don't know how rust versioning works. I'm using the [semver](https://semver.org/) mental model.

chancehudson:
Hey @jan-ferdinand thinking about this more it would be very useful if the triton-vm crate provided a standard way to serialize a proof. What is your opinion of pulling the serialization logic into triton-vm itself?

jan-ferdinand:
> it would be useful if Triton VM provided a standard way to serialize a proof

I think this makes sense, yes.

erikerlandson:
Is this PR still active?

jan-ferdinand:
Not currently. I still think it's nice to have a CLI for proving & verifying. I got sidetracked during the (de)serialization discussion somewhere and honestly forgot to come back to it. Skimming it now, I'm still not sure how best to go about things.

jan-ferdinand:
I made a [stand-alone CLI](https://github.com/TritonVM/triton-cli) for running and proving Triton programs and verifying those proofs. You can install it directly through `cargo install triton-cli`.

That CLI does not have all the same features as the CLI of this PR; in particular, when verifying, no information about the proof is communicated beyond verification success or failure.

However,
- public input is supported, both “inline” and from a file,
- secret input is supported through a json-serialized [`NonDeterminism`](https://docs.rs/triton-vm/0.48.0/triton_vm/vm/struct.NonDeterminism.html), and
- a json-serialized [`VMState`](https://docs.rs/triton-vm/0.48.0/triton_vm/vm/struct.VMState.html) can be used for both running and proving.

I have not solved the UX problem that verifying proofs for an incompatible version of Triton VM will fail without a clear reason beyond.

PR #314, opened 2024-08-09
feat: Dynamic Address for Out-of-Domain Row in TASM AIR Evaluation

aszepieniec:
The current TASM AIR evaluation code is very efficient but has a major drawback: the location of the inputs to the circuit (*i.e.*, the {current, next} {main, aux} out-of-domain (OOD) rows) are statically defined and hence must be known at compile time. As a result, one runs into complications when verifying multiple STARK proofs in a single TASM program; and verifying a dynamic number of them is all but impossible. Moreover, the static location of the OOD rows are baked into the computer program, meaning that even if you do compile a TASM program that verifies multiple proofs, you get different programs (and different digests!) depending on how many proofs you actually need to verify.

This PR adds a second function for evaluating the AIR in tasm. It is a dynamic counterpart to the static one that's already there.

The new function *does* take a memory layout object, just like the old function does. But the difference is that this new memory layout object only specifies where to find the challenges and where to find free memory. The locations of the OOD rows is given as arguments on the stack. The method then stores these locations in the first few words of the free memory region. Whenever an input needs to be read, the relevant pointer-pointer is read and dereferenced twice -- the second time with the appropriate offset. (In contrast, the static code derefences a regular pointer once.)

In more detail, here are the changes introduced by this PR:
 - Update spec and spec-verifying tests.
 - Rename current tasm code to reflect static nature.
 - Add dynamic tasm code emitter.
 - Integrate dynamic tasm evaluator into existing tests.
 - Restructure: file `air.rs` and directory `air` contain logic related to evaluating AIR.


triton-vm/src/air.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+    }
+
+    #[proptest]
+    fn triton_assembly_constraint_evaluators_are_straight_line_and_does_not_halt(
+        #[strategy(arb())] static_memory_layout: StaticTasmConstraintEvaluationMemoryLayout,
+        #[strategy(arb())] dynamic_memory_layout: DynamicTasmConstraintEvaluationMemoryLayout,
+    ) {
+        type I = AnInstruction<String>;
+        let is_legal = |i| !matches!(i, I::Call(_) | I::Return | I::Recurse | I::Skiz | I::Halt);
```
jan-ferdinand, line 256: My oversight: `RecurseOrReturn` shoud be on this list.

triton-vm/src/air/memory_layout.rs
```diff
@@ -0,0 +1,212 @@
...
+///
+/// [tasm_air]: tasm_air_constraints::dynamic_air_constraint_evaluation_tasm
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Arbitrary)]
+pub struct DynamicTasmConstraintEvaluationMemoryLayout {
+    /// Pointer to a region of memory that is reserved for a) pointers to {current,
+    /// next} {main, aux} rows, and b) intermediate values in the course of constraint
+    /// evaluation. The size of the region must be at least [`MEM_PAGE_SIZE`][mem_page_size]
+    /// [`BFieldElement`]s.
+    ///
+    /// [mem_page_size]: TasmConstraintEvaluationMemoryLayout::MEM_PAGE_SIZE
```
jan-ferdinand, line 19:
```suggestion
    /// [mem_page_size]: IntegralMemoryLayout::MEM_PAGE_SIZE
```

And also below.

triton-vm/src/table.rs
```diff
@@ -32,7 +30,6 @@ pub mod program_table;
 pub mod ram_table;
 pub mod table_column;
 #[rustfmt::skip]
-pub mod tasm_air_constraints;
```
jan-ferdinand, line 35: The `rustfmt::skip` should be moved along with `tasm_air_constraints`. Now, it applies to `u32_table`, which is probably unintended.

jan-ferdinand (changes requested):
It seems like there is a huge amount of copied code in the new dynamic generator. This does not feel necessary. Looks fine other than that.

PR #316, opened 2024-08-11
add comments for fill_program_hash_trace function

Einstellung:
Add comments to the function to make the code easier to understand for the reader

### DIFF
--- a/triton-vm/src/aet.rs
+++ b/triton-vm/src/aet.rs
@@ -166,12 +166,14 @@ impl AlgebraicExecutionTrace {
     /// Hash the program and record the entire Sponge's trace for program attestation.
     fn fill_program_hash_trace(&mut self) {
         let padded_program = Self::hash_input_pad_program(&self.program);
+        // sponge init
         let mut program_sponge = Tip5::init();
         for chunk in padded_program.chunks(Tip5::RATE) {
             program_sponge.state[..Tip5::RATE]
                 .iter_mut()
                 .zip_eq(chunk)
                 .for_each(|(sponge_state_elem, &absorb_elem)| *sponge_state_elem = absorb_elem);
+            // sponge absorb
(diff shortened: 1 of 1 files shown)


jan-ferdinand:
Thanks for the PR. :blush: I would argue that Sponge absorption is more than just application of the permutation to the current state; replacing part of the Sponge's state with the chunk of size `rate` is already part of absorption. That said, comments or better variable names can indeed contribute to readability here. :+1: How about:

``` rust
fn fill_program_hash_trace(&mut self) {
    let padded_program = Self::hash_input_pad_program(&self.program);
    let mut sponge = Tip5::init(); // I think the names here make it clear that this is initialization
    for chunk_to_absorb in padded_program.chunks(Tip5::RATE) { // renaming from `chunk`
        sponge.state[..Tip5::RATE]
            .iter_mut()
            .zip_eq(chunk_to_absorb)
            .for_each(|(sponge_state_elem, &absorb_elem)| *sponge_state_elem = absorb_elem);
        let hash_trace = sponge.trace();

        self.increase_lookup_multiplicities(hash_trace);
        let trace_addendum = HashTable::trace_to_table_rows(hash_trace);
        self.program_hash_trace
            .append(Axis(0), trace_addendum.view())
            .expect("shapes must be identical");
    }
	// …
}
```

Einstellung:
Yes, you are right. The `for chunk...` loop is also part of the sponge absorption process. Initially, I didn't have a suitable place to commit, so i just committed to the final line. Your decision to change the variable name is better than my initial approach and makes it easier for readers to understand😄. I agree with your changes and will close this pull request. I will follow your changes and then submit a new pull request.

Einstellung:
I think keep `program_sponge` is better than using `sponge` because it makes it clear to the reader that this sponge process is specifically for program data.

PR #325, opened 2024-09-12
Thv/degree lowering hardening

Sword-Smith:
Add tests that degree lowering and code generator for constraint evaluation is deterministic and does not change unintentionally.

### DIFF
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -1968,7 +1968,7 @@ mod tests {
         let CircuitExpression::BinOp(BinOp::Add, variable, neg_expression) =
             &substitution_constraint.circuit.as_ref().borrow().expression
         else {
-            unreachable!();
+            panic!();
         };
         let extra_input =
             match &neg_expression.as_ref().borrow().expression {
@@ -1978,7 +1978,7 @@ mod tests {
                 CircuitExpression::BConst(c) => -c.lift(),
                 CircuitExpression::XConst(c) => -*c,
                 _ => {
-                    unreachable!()
+                    panic!()
                 }
             };
         if variable.borrow().evaluates_to_base_element() {
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -801,7 +801,7 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
         let new_variable = builder.input(new_input_indicator);
 
         // Make all descendants of the chosen node point to the new variable instead
-        builder.redirect_pointers(chosen_node_id, new_variable.clone());
+        builder.redirect_all_references_to_node(chosen_node_id, new_variable.clone());
 
         // Treat roots of the multicircuit explicitly.
         for circuit in multicircuit.iter_mut() {
@@ -1039,7 +1039,7 @@ impl<II: InputIndicator> ConstraintCircuitBuilder<II> {
     ///
     /// A circuit's root node cannot be substituted with this method. Manual care
     /// must be taken to update the root node if necessary.
-    fn redirect_pointers(&self, old_id: usize, new: ConstraintCircuitMonad<II>) {
+    fn redirect_all_references_to_node(&self, old_id: usize, new: ConstraintCircuitMonad<II>) {
         self.all_nodes.borrow_mut().remove(&old_id);
         for node in self.all_nodes.borrow_mut().values_mut() {
             let node_expression = &mut node.circuit.borrow_mut().expression;
@@ -1486,7 +1486,10 @@ mod tests {
         assert!(root_2.contains(&substitute_me));
 
         let new_variable = x(3);
-        builder.redirect_pointers(substitute_me.circuit.borrow().id, new_variable.clone());
+        builder.redirect_all_references_to_node(
+            substitute_me.circuit.borrow().id,
+            new_variable.clone(),
+        );
 
         assert!(!root_0.contains(&substitute_me));
         assert!(!root_1.contains(&substitute_me));
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -707,6 +707,20 @@ impl<II: InputIndicator> Sum for ConstraintCircuitMonad<II> {
     }
 }
 
+struct EvolvingMainConstraintsNumber(usize);
+impl From<EvolvingMainConstraintsNumber> for usize {
+    fn from(value: EvolvingMainConstraintsNumber) -> Self {
+        value.0
+    }
+}
+
+struct EvolvingAuxConstraintsNumber(usize);
+impl From<EvolvingAuxConstraintsNumber> for usize {
+    fn from(value: EvolvingAuxConstraintsNumber) -> Self {
+        value.0
+    }
+}
+
 impl<II: InputIndicator> ConstraintCircuitMonad<II> {
     /// Unwrap a ConstraintCircuitMonad to reveal its inner ConstraintCircuit
     pub fn consume(&self) -> ConstraintCircuit<II> {
@@ -761,8 +775,8 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
                 multicircuit,
                 info,
                 chosen_node_id,
-                main_constraints.len(),
-                aux_constraints.len(),
+                EvolvingMainConstraintsNumber(main_constraints.len()),
+                EvolvingAuxConstraintsNumber(aux_constraints.len()),
             );
 
             if new_constraint.circuit.borrow().evaluates_to_base_element() {
@@ -785,8 +799,8 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
         multicircuit: &mut [Self],
         info: DegreeLoweringInfo,
         chosen_node_id: usize,
-        new_main_constraints_count: usize,
-        new_aux_constraints_count: usize,
+        new_main_constraints_count: EvolvingMainConstraintsNumber,
+        new_aux_constraints_count: EvolvingAuxConstraintsNumber,
     ) -> ConstraintCircuitMonad<II> {
         let builder = multicircuit[0].builder.clone();
 
@@ -794,10 +808,10 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
         let chosen_node = builder.all_nodes.borrow()[&chosen_node_id].clone();
         let chosen_node_is_main_col = chosen_node.circuit.borrow().evaluates_to_base_element();
         let new_input_indicator = if chosen_node_is_main_col {
-            let new_main_col_idx = info.num_main_cols + new_main_constraints_count;
+            let new_main_col_idx = info.num_main_cols + usize::from(new_main_constraints_count);
             II::main_table_input(new_main_col_idx)
         } else {
-            let new_aux_col_idx = info.num_aux_cols + new_aux_constraints_count;
+            let new_aux_col_idx = info.num_aux_cols + usize::from(new_aux_constraints_count);
             II::aux_table_input(new_aux_col_idx)
         };
         let new_variable = builder.input(new_input_indicator);
@@ -1951,8 +1965,8 @@ mod tests {
             &mut multicircuit_monad,
             degree_lowering_info,
             substitution_node_id,
-            0,
-            0,
+            EvolvingMainConstraintsNumber(0),
+            EvolvingAuxConstraintsNumber(0),
         );
 
         // extract substituted constraint
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -276,11 +276,11 @@ impl InputIndicator for DualRowIndicator {
 /// - challenges, _i.e._, (pseudo-)random values sampled through the Fiat-Shamir
 ///   heuristic.
 ///
-/// An internal node, representing some binary operation, is either addition, multiplication.
-/// The left and right children of the node are the operands of the binary operation.
-/// The left and right children are not themselves `CircuitExpression`s, but rather
-/// [`ConstraintCircuit`]s, which is a wrapper around `CircuitExpression` that manages
-/// additional bookkeeping information.
+/// An internal node, representing some binary operation, is either addition or
+/// multiplication. The left and right children of the node are the operands of
+/// the binary operation. The left and right children are not themselves `CircuitExpression`s,
+/// but rather [`ConstraintCircuit`]s, which is a wrapper around `CircuitExpression`
+/// that manages additional bookkeeping information.
 #[derive(Debug, Clone)]
 pub enum CircuitExpression<II: InputIndicator> {
     BConst(BFieldElement),
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -1230,53 +1230,53 @@ mod tests {
             self_expression.contains(other_expression)
         }
 
+        /// Produces an iter over all nodes in the multicircuit, if it is non-empty.
+        ///
+        /// Helper function for counting the number of nodes of a specific type.
+        fn iter_nodes(
+            constraints: &[Self],
+        ) -> std::vec::IntoIter<(usize, ConstraintCircuitMonad<II>)> {
+            if let Some(first) = constraints.first() {
+                first
+                    .builder
+                    .all_nodes
+                    .borrow()
+                    .iter()
+                    .map(|(n, m)| (*n, m.clone()))
+                    .collect_vec()
+                    .into_iter()
+            } else {
+                vec![].into_iter()
+            }
+        }
+
         /// Counts the number of inputs from the main table
         fn num_nodes(constraints: &[Self]) -> usize {
-            constraints
-                .first()
-                .unwrap()
-                .builder
-                .all_nodes
-                .borrow()
-                .len()
+            Self::iter_nodes(constraints).count()
+        }
+
+        /// Determine if the constraint circuit monad corresponds to a main table
+        /// column.
+        fn is_main_table_column(&self) -> bool {
+            if let CircuitExpression::Input(ii) = self.circuit.as_ref().borrow().expression {
+                ii.is_main_table_column()
+            } else {
+                false
+            }
         }
 
         /// Counts the number of inputs from the main table
         fn num_main_inputs(constraints: &[Self]) -> usize {
-            constraints
-                .first()
-                .unwrap()
-                .builder
-                .all_nodes
-                .borrow()
-                .iter()
-                .filter(|(_, cc)| {
-                    if let CircuitExpression::Input(ii) = cc.circuit.as_ref().borrow().expression {
-                        ii.is_main_table_column()
-                    } else {
-                        false
-                    }
-                })
+            Self::iter_nodes(constraints)
+                .filter(|(_, cc)| cc.is_main_table_column())
                 .filter(|(_, cc)| cc.circuit.borrow().evaluates_to_base_element())
                 .count()
         }
 
         /// Counts the number of inputs from the aux table
         fn num_aux_inputs(constraints: &[Self]) -> usize {
-            constraints
-                .first()
-                .unwrap()
-                .builder
-                .all_nodes
-                .borrow()
-                .iter()
-                .filter(|(_, cc)| {
-                    if let CircuitExpression::Input(ii) = cc.circuit.as_ref().borrow().expression {
-                        !ii.is_main_table_column()
-                    } else {
-                        false
-                    }
-                })
+            Self::iter_nodes(constraints)
+                .filter(|(_, cc)| !cc.is_main_table_column())
                 .count()
         }
 
@@ -1287,13 +1287,7 @@ mod tests {
 
         /// Counts the number of challenges
         fn num_challenges(constraints: &[Self]) -> usize {
-            constraints
-                .first()
-                .unwrap()
-                .builder
-                .all_nodes
-                .borrow()
-                .iter()
+            Self::iter_nodes(constraints)
                 .filter(|(_, cc)| {
                     matches!(
                         cc.circuit.as_ref().borrow().expression,
@@ -1305,13 +1299,7 @@ mod tests {
 
         /// Counts the number of `BinOp`s
         fn num_binops(constraints: &[Self]) -> usize {
-            constraints
-                .first()
-                .unwrap()
-                .builder
-                .all_nodes
-                .borrow()
-                .iter()
+            Self::iter_nodes(constraints)
                 .filter(|(_, cc)| {
                     matches!(
                         cc.circuit.as_ref().borrow().expression,
@@ -1323,13 +1311,7 @@ mod tests {
 
         /// Counts the number of BFE constants
         fn num_bfield_constants(constraints: &[Self]) -> usize {
-            constraints
-                .first()
-                .unwrap()
-                .builder
-                .all_nodes
-                .borrow()
-                .iter()
+            Self::iter_nodes(constraints)
                 .filter(|(_, cc)| {
                     matches!(
                         cc.circuit.as_ref().borrow().expression,
@@ -1341,13 +1323,7 @@ mod tests {
 
         /// Counts the number of XFE constants
         fn num_xfield_constants(constraints: &[Self]) -> usize {
-            constraints
-                .first()
-                .unwrap()
-                .builder
-                .all_nodes
-                .borrow()
-                .iter()
+            Self::iter_nodes(constraints)
                 .filter(|(_, cc)| {
                     matches!(
                         cc.circuit.as_ref().borrow().expression,
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -1103,7 +1103,8 @@ mod tests {
     use std::hash::Hasher;
 
     use itertools::Itertools;
-    use ndarray::{Array2, Axis};
+    use ndarray::Array2;
+    use ndarray::Axis;
     use proptest::arbitrary::Arbitrary;
     use proptest::collection::vec;
     use proptest::prelude::*;
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -1834,11 +1834,6 @@ mod tests {
         );
 
         prop_assert_eq!(10, ConstraintCircuitMonad::num_inputs(&multicircuit_monad));
-        prop_assert_eq!(
-            10,
-            ConstraintCircuitMonad::num_bfield_constants(&multicircuit_monad)
-                + ConstraintCircuitMonad::num_xfield_constants(&multicircuit_monad)
-        );
     }
 
     /// Test the completeness and soundness of the `apply_substitution` function,
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -1276,7 +1276,13 @@ mod tests {
         /// The number of inputs from the aux table
         fn num_aux_inputs(constraints: &[Self]) -> usize {
             Self::iter_nodes(constraints)
-                .filter(|(_, cc)| !cc.is_main_table_column())
+                .filter(|(_, cc)| {
+                    if let CircuitExpression::Input(ii) = cc.circuit.as_ref().borrow().expression {
+                        !ii.is_main_table_column()
+                    } else {
+                        false
+                    }
+                })
                 .count()
         }
 
@@ -1716,13 +1722,13 @@ mod tests {
         num_inputs: usize,
         num_challenges: usize,
         num_constants: usize,
-        num_binops: usize,
+        num_operations: usize,
         num_outputs: usize,
     ) -> BoxedStrategy<Vec<ConstraintCircuitMonad<II>>> {
         (
             vec(CircuitInputType::arbitrary(), num_inputs),
             vec(CircuitConstantType::arbitrary(), num_constants),
-            vec(CircuitOperationChoice::arbitrary(), num_binops),
+            vec(CircuitOperationChoice::arbitrary(), num_operations),
             vec(arb::<usize>(), num_outputs),
         )
             .prop_map(move |(inputs, constants, operations, outputs)| {
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -1856,7 +1856,7 @@ mod tests {
     /// is not the zero vector (with high probability) and so the given input is
     /// *not* a satisfying assignment (with high probability). However, the
     /// circuit can be extended by way of thought experiment into one that subtracts
-    /// a fixed constant from the original output. For the right choice of substrahend,
+    /// a fixed constant from the original output. For the right choice of subtrahend,
     /// the random input now *is* a satisfying assignment to the circuit.
     ///
     /// Specifically, let `input` denote the original (before degree lowering) input,
@@ -1884,25 +1884,25 @@ mod tests {
     ///
     /// This description makes use of the following commutative diagram.
     ///
-    /// ```notest
-    ///          C ---- degree-lowering -----> C || k
-    ///          |                               |
-    /// subtract |                      subtract |
-    ///    fixed |                         fixed |
-    ///   output |                        output |
-    ///          |                               |
+    /// ```text
+    ///          C ───── degree-lowering ────> C || k
+    ///          │                               │
+    /// subtract │                      subtract │
+    ///    fixed │                         fixed │
+    ///   output │                        output │
+    ///          │                               │
     ///          v                               v
-    ///          C* --- degree-lowering ----> C* || k
+    ///          C* ─── degree-lowering ────> C* || k
     /// ```
     ///
     /// The point of this elaboration is that in this particular case, testing completeness
     /// and soundness require the same code path. If `input` and `input || some_expr(input)`
     /// work for circuits before and after degree lowering, this fact establishes
     /// both completeness and soundness simultaneously.
-    ///
-    /// Shrinking on this test is disabled because we noticed some weird ass behavior.
-    /// In short, shrinking does not play ball with the arbitrary circuit generator;
-    /// it seems to make the generated circuits *more* complex, not less so.
+    //
+    // Shrinking on this test is disabled because we noticed some weird ass behavior.
+    // In short, shrinking does not play ball with the arbitrary circuit generator;
+    // it seems to make the generated circuits *more* complex, not less so.
     #[proptest(cases = 1000, max_shrink_iters = 0)]
     fn node_substitution_is_complete_and_sound(
         #[strategy(arbitrary_circuit_monad(10, 10, 10, 200, 10))] mut multicircuit_monad: Vec<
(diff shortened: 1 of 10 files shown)


triton-constraint-circuit/src/lib.rs
```diff
@@ -775,6 +763,41 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
         (main_constraints, aux_constraints)
     }
 
+    fn substitute_node_and_produce_constraint(
```
jan-ferdinand, line 766: A function that does two things is bad design.[^and] That said, the entire degree lowering pipeline violates good software engineering practices in various places and needs an overhaul. If it is possible to have this function do one thing only, that design is preferred. If this is too difficult given the context the function lives in, that design flaw will have to be eliminated with the general overhaul of constraint building and degree lowering.

[^and]: Function names containing `_and_` are a red flag for this.

triton-constraint-circuit/src/lib.rs
```diff
@@ -1390,6 +1413,57 @@ mod tests {
...
+    #[test]
+    fn pick_node_to_substitute_is_deterministic() {
+        let multicircuit = circuit_with_multiple_options_for_degree_lowering_to_degree_2();
+        let first_node_id = ConstraintCircuitMonad::pick_node_to_substitute(&multicircuit, 2);
+
+        for _ in 0..20 {
+            let node_id_again = ConstraintCircuitMonad::pick_node_to_substitute(&multicircuit, 2);
+            assert_eq!(first_node_id, node_id_again);
+        }
+    }
```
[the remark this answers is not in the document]
aszepieniec: Your suggestion tests that every consecutive disjoint pair of invocations to `pick_node_to_substitute` agrees. The original unit test tests that 21 consecutive invocations all agree, which is a stronger thing.

triton-vm/src/table/master_table.rs
```diff
@@ -2069,4 +2071,115 @@ mod tests {
         let expected_digest = Tip5::hash_varlen(&elements);
         prop_assert_eq!(expected_digest, pending_absorb_digest);
     }
+
+    /// Test whether the AIR constraint evaluators are the same between
+    ///  (a) the time when this test was written or last updated; and
+    ///  (b) the time when the test is being executed.
+    ///
```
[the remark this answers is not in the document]
aszepieniec: IMHO the first operand (a) of the comparison is left underconstrained by the function name, and is thus ambiguous.

triton-vm/src/table/master_table.rs
```diff
@@ -2069,4 +2071,115 @@ mod tests {
...
+    /// This test catches (whp) unintended changes, whether due to
+    /// nondeterminisms (on a single machine or across various machines) or due
+    /// to changes to the definitions of the constraints. If the change to the
+    /// constraints was intentional, this test should fail until the expected
+    /// value in the assert-statement is updated.
```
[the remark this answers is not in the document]
aszepieniec: This suggestion misses the opportunity to reassure the programmer that there's nothing wrong with his code.

triton-vm/src/table/master_table.rs
```diff
@@ -2069,4 +2071,115 @@ mod tests {
...
+    #[test]
+    fn air_constraints_evaluators_have_not_changed() {
+        let seed: [u8; 32] = [
+            193, 123, 173, 34, 45, 172, 73, 224, 156, 208, 121, 29, 245, 142, 215, 124, 188, 114,
+            203, 148, 186, 245, 224, 76, 142, 129, 82, 225, 81, 28, 216, 183,
+        ];
+
+        let mut rng: StdRng = SeedableRng::from_seed(seed);
```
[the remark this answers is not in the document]
aszepieniec: The given seed was produced by `thread_rng()` and therefore guaranteed to be random. In contrast, `0` doesn't seem very random. Are you sure you produced it from `thread_rng()` or something similar?

Moreover, applying this suggestion causes the test to fail, because the expected value comes from a different seed.

triton-constraint-circuit/src/lib.rs
aszepieniec: This function has been renamed to `apply_substitution`.

aszepieniec:
Something funky is happening with the test `table::master_table::tests::air_constraints_evaluators_have_not_changed` which I can't explain. CI is failing with
```
expected: (03756754445351585926·x² + 06769459618545093804·x + 09140558386905394900)
observed: (14714205515170095309·x² + 13661591558076455036·x + 14115119063754412580)
```
whereas on my desktop computer I get
```
expected: (03756754445351585926·x² + 06769459618545093804·x + 09140558386905394900)
observed: (03346962278739234229·x² + 12572006219868624302·x + 10286558549229477187)
```
and *both* are obviously different from my laptop which produced the expected value. Any ideas as to what might cause this?

Sword-Smith:
On my desktop at home, I get:
```
expected: (03756754445351585926·x² + 06769459618545093804·x + 09140558386905394900)
observed: (14714205515170095309·x² + 13661591558076455036·x + 14115119063754412580)
```

Sword-Smith:
Would it make sense to also add a proptest like `node_substitution_is_complete_and_sound` for a `DualRowIndicator` type?

aszepieniec:
Correction: on my laptop I get this failure:
```
expected: (03756754445351585926·x² + 06769459618545093804·x + 09140558386905394900)
observed: (14714205515170095309·x² + 13661591558076455036·x + 14115119063754412580)
```
which agrees with CI and @Sword-Smith's machine.

Mysteries:
 - why did my desktop produce a different result?
 - why did my laptop produce a different result in the past?

aszepieniec:
For the record, applying just the first commit (9e5db866063a6527366be62616611d74db1ef416) of setting `opt-level=3` generates this failure:
```
expected: (03756754445351585926·x² + 06769459618545093804·x + 09140558386905394900)
observed: (03346962278739234229·x² + 12572006219868624302·x + 10286558549229477187)
```
which agrees with the failure before setting the opt-level. So this compiler optimization does not seem to change anything.

triton-vm/src/table/master_table.rs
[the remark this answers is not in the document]
aszepieniec: For one isolated test, there is no practical difference and your suggestion is simpler.

In the context of many derandomized probabilistic tests, you reduce the quality of the entire test suite by always using the same seed. You want your tests to be independent of each other and always sampling the same indeterminate (for instance) correlates them. 

In light of that consideration, I would suggest that it is good practice to always use independent random seeds in derandomized probabilistic tests. (Even if that complicates the code, and even if it is only one test case.)

aszepieniec:
@jan-ferdinand: All checks pass. Any further requested changes?

triton-constraint-circuit/src/lib.rs
```diff
@@ -775,6 +763,41 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
...
+        } else {
+            let new_aux_col_idx = info.num_aux_cols + aux_constraints_count;
+            II::aux_table_input(new_aux_col_idx)
+        };
+        let new_variable = builder.input(new_input_indicator);
+
+        // Substitute the chosen circuit with the new variable.
+        builder.substitute(chosen_node_id, new_variable.clone());
+
+        // Treat roots of the multicircuit explicitly.
```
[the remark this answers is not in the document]
aszepieniec: Renamed to `redirect_pointers` which more accurately reflects what's going on underneath.

aszepieniec:
BTW: I can confirm that all tests pass on my desktop machine as well.

triton-vm/src/table/master_table.rs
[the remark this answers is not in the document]
jan-ferdinand: Those arguments are sound. They do not preclude using `let mut rng = StdRng::seed_from_u64(/* independent random seed */);`

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1696,328 @@ mod tests {
...
+            vec(arb::<usize>(), num_outputs),
+        )
+            .prop_map(move |(inputs, constants, operations, outputs)| {
+                let builder = ConstraintCircuitBuilder::<II>::new();
+
+                assert_eq!(0, *builder.id_counter.as_ref().borrow());
+                assert!(
+                    builder.all_nodes.borrow().is_empty(),
+                    "fresh hashmap should be empty"
+                );
```
jan-ferdinand, line 1741: Maybe add a comment that these asserts were added because there exists a suspicion that shrinking re-uses objects?

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1696,328 @@ mod tests {
...
+        num_constants: usize,
+        num_nodes: usize,
+        num_outputs: usize,
+    ) -> BoxedStrategy<Vec<ConstraintCircuitMonad<II>>> {
+        (
+            vec(CircuitInputType::arbitrary(), num_inputs),
+            vec(CircuitConstantType::arbitrary(), num_constants),
+            vec(
+                CircuitOperationChoice::arbitrary(),
+                num_nodes - num_inputs - num_constants - num_outputs,
```
jan-ferdinand, line 1730: Would changing `num_nodes` into `num_additional_nodes` and changing this line into `num_inputs + num_constants + num_outputs + num_additional_nodes` simplify the relation between the input parameters? Or are there additional, hidden requirements about their relation?

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1696,328 @@ mod tests {
...
+                        CircuitConstantType::Base(bfe) => {
+                            let node = builder.b_constant(bfe);
+                            all_nodes.push(node);
+                        }
+                        CircuitConstantType::Extension(xfe) => {
+                            let node = builder.x_constant(xfe);
+                            all_nodes.push(node);
+                        }
+                    }
+                }
```
jan-ferdinand, line 1779: I understand why you don't want to factor out `all_nodes.push(…)` in the inputs loop. Here, there are no difficulties with incrementing state variables.

```suggestion
for constant in constants {
    let node = match constant {
        CircuitConstantType::Base(bfe) => builder.b_constant(bfe),
        CircuitConstantType::Extension(xfe) => builder.x_constant(xfe),
    };
    all_nodes.push(node);
}
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1696,328 @@ mod tests {
...
+
+                if !all_nodes.is_empty() {
```
jan-ferdinand, line 1781: Would an early return make sense?

```rust
if all_nodes.is_empty() {
	return vec![];
}
```

Now, the `operations` and the `output` loop can both assume that `all_nodes` is non-empty.

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1696,328 @@ mod tests {
...
+                                let lhs_index = lhs % all_nodes.len();
+                                let rhs_index = rhs % all_nodes.len();
+
+                                let lhs_node = all_nodes[lhs_index].clone();
+                                let rhs_node = all_nodes[rhs_index].clone();
+
+                                let node = lhs_node * rhs_node;
+                                all_nodes.push(node);
+                            }
+                        }
```
jan-ferdinand, line 1804: Common code can be factored out.

```suggestion
                        let (lhs, rhs) = match operation {
                            CircuitOperationChoice::Add(lhs, rhs) => (lhs, rhs),
                            CircuitOperationChoice::Mul(lhs, rhs) => (lhs, rhs),
                        };
    
                        let lhs_index = lhs % all_nodes.len();
                        let rhs_index = rhs % all_nodes.len();
    
                        let lhs_node = all_nodes[lhs_index].clone();
                        let rhs_node = all_nodes[rhs_index].clone();
    
                        let node = match operation {
                            CircuitOperationChoice::Add(_, _) => lhs_node + rhs_node,
                            CircuitOperationChoice::Mul(_, _) => lhs_node * rhs_node,
                        };
                        all_nodes.push(node);
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1696,328 @@ mod tests {
...
+    ///
+    /// Shrinking on this test is disabled because we noticed some weird ass behavior.
+    /// In short, shrinking does not play ball with the arbitrary circuit generator;
+    /// it seems to make the generated circuits *more* complex, not less so.
+    #[proptest(cases = 1000, max_shrink_iters = 0)]
+    fn node_substitution_is_complete_and_sound(
+        #[strategy(arbitrary_circuit_monad(10, 10, 10, 200, 10))] mut multicircuit_monad: Vec<
+            ConstraintCircuitMonad<SingleRowIndicator>,
+        >,
+        #[strategy(vec(arb::<BFieldElement>(), ConstraintCircuitMonad::num_main_inputs(&#multicircuit_monad)))]
```
jan-ferdinand, line 1902: If you let type inference do its thing this line will be easer to read, maintain, and also respect the max line width.

```suggestion
        #[strategy(vec(arb(), ConstraintCircuitMonad::num_main_inputs(&#multicircuit_monad)))]
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1696,328 @@ mod tests {
...
+        #[filter(!#aux_input.is_empty())]
+        aux_input: Vec<XFieldElement>,
+        #[strategy(vec(arb::<XFieldElement>(), ConstraintCircuitMonad::num_challenges(&#multicircuit_monad)))]
+        challenges: Vec<XFieldElement>,
+        #[strategy(arb())] substitution_node_index: usize,
+    ) {
+        let mut main_input = Array2::from_shape_vec((1, main_input.len()), main_input).unwrap();
+        let mut aux_input = Array2::from_shape_vec((1, aux_input.len()), aux_input).unwrap();
+
+        // compute circuit output before degree-lowering
```
jan-ferdinand, line 1915: The variable name implies the comment.

jan-ferdinand (changes requested):
Pending: proptest “node_substitution_is_complete_and_sound”

triton-constraint-circuit/src/lib.rs
```diff
@@ -775,6 +763,41 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
         (main_constraints, aux_constraints)
     }
 
+    fn substitute_node_and_produce_constraint(
+        multicircuit: &mut [Self],
+        info: DegreeLoweringInfo,
+        chosen_node_id: usize,
+        main_constraints_count: usize,
+        aux_constraints_count: usize,
```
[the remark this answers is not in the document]
aszepieniec: f9f83dc675016182b4f3bbe2a57f8fcbb9e8cdc6 Introduces type aliases `EvolvingMainConstraintsNumber` and `EvolvingAuxConstraintsNumber` for the second two counts.

Personally, I think this change degrades code quality. Having the commit as part of the PR is the fastest way forward, because it either earns your checkmark directly or else contributes to convincing you to reject it.

My position, in case you are interested:
 - Types used once are mis-used.
 - The function is called in one place only, not counting the test. It is conceptually part of that same function; the only reason why it is factored out is to test it in isolation.
 - The counts serve to derive indices of new columns and are thus inherently `usize`s. Declaring another type for these variables suggests a different context applies, but I don't think it does. The reader's attention is diverted towards finding this missing context.
 - Messing up the order of the arguments is still possible. (If due to too-weak introduced types, please show by example how it ought to be done.)
 - Why not declare a new type for every argument in every function interface? I do think that's the logical conclusion of your train of thought; if not I wonder where it stops.
 - All of the above being said, I do think that `NodeId` could be a useful type but a) that doesn't resolve the problem completely; and b) that refactor is out of the scope of this PR.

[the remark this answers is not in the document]
jan-ferdinand: [Type aliases](https://doc.rust-lang.org/book/ch19-04-advanced-types.html#creating-type-synonyms-with-type-aliases) don't do what you suggest they do in 66b8225 – that's [newtypes](https://doc.rust-lang.org/book/ch19-04-advanced-types.html#using-the-newtype-pattern-for-type-safety-and-abstraction).

Taking position:

> Types used once are mis-used.

I see help from the type system to write correct programs as beneficial in all cases.[^counter]

> The function is called in one place only […].

I don't see how this justifies bad design.

>  […] the […] reason why it is factored out is to test it in isolation.

Which is laudable and a decision I fully support.

> The counts […] are […] inherently `usize`s.

Consider:

```rust
let _ = PrintConfig::new(true, true, false);
let _ = PrintConfig::new(Sided::Single, Colored::Colored, PrintSpeed::Slow);
```

Even though (for arguments sake) all these enums have only two variants, and are thus inherently `bool`s, I strongly prefer the second version.[^better] In my opinion, the argument extends to numerical types.

> Why not declare a new type for every argument in every function interface?

The concrete contract of a function as well the number of arguments it takes make a huge difference. In `my_vec.remove(i)`, there's no room for screwing up the argument order.[^tech] The same holds for `my_vec.swap(i, j)`, and so having two arguments of the same type does not influence the method's usability. I don't have a hard-and-fast rule at hand as to where it is beneficial to introduce new types.

> I do think that `NodeId` could be a useful type

Sounds like a good idea. I agree with it being out of scope for this PR.

---

All that said, I think testability completely justifies factoring out the function. I also think that you have inherited design decisions (and, worse, accidental design) that would require a ton of work to improve.[^ind] We can write the argument sublist `usize, usize, usize` off as being part of overall improvable design, improving which is left as future work. One might argue that this constitutes building up technical debt,[^debt] but I think an equally valid argument is that factoring out the method only highlights existing technical debt.

In my opinion, introducing a type like `NumConstraints` slightly mitigates the issues I have with the same-type arguments. Fully resolving all pain points I see would probably require a more fundamental redesign.

[^better]: Other designs than a function `new()` taking many arguments should be considered were `PrintConfig` actually being added. The types' names would need a rework, too.
[^tech]: Technically, it's possible: `Vec::remove(i, &mut my_vec)`. However, (a) the type checker will stop you, and (b) why would you write that, apart from trying to prove a point, which, fair enough.
[^counter]: Yes, I'm looking forward to your counter argument after having made such an absolute claim.
[^ind]:  One strong indicator of this is that the two arguments in question are essentially internal tracking information. The function should not have to rely on externally supplied information for this.
[^debt]:  I'm sure you won't be surprised to hear that I'm not a fan of building up technical debt, and even less of doing it consciously.

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+                constraint.circuit.borrow().evaluate(
+                    main_input.view(),
+                    aux_input.view(),
+                    &challenges,
+                )
+            })
+            .collect_vec();
+
+        // apply one step of degree-lowering
+        let num_nodes = multicircuit_monad[0].builder.all_nodes.borrow().len();
```
jan-ferdinand, line 1925: You introduced a helper function for this:

```suggestion
        let num_nodes = ConstraintCircuitMonad::num_nodes(&multicircuit_monad);
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+                .builder
+                .all_nodes
+                .borrow()
+                .iter()
+                .nth((substitution_node_index + j) % num_nodes)
+                .map(|(i, n)| (*i, n.clone()))
+                .unwrap();
+            j += 1;
+        }
+        prop_assert_ne!(j, num_nodes, "no suitable nodes to substitute");
```
jan-ferdinand, line 1946: The following removes the duplicated code. Since `substituted` is no longer mutable, the compiler correctly detects that it is never used.

```suggestion
        let (&substitution_node_id, _) = multicircuit_monad[0]
            .builder
            .all_nodes
            .borrow()
            .iter()
            .cycle()
            // `% num_nodes` is not even necessary due to the `cycle()`
            // but I think it communicates intent. Happy with either.
            // This comment is only intended for the review, not the code.
            .skip(substitution_node_index % num_nodes)
            .take(num_nodes)
            .find(|(_, monad)| !monad.circuit.borrow().is_zero())
            .expect("no suitable nodes to substitute");
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+        let output_before_lowering = multicircuit_monad
+            .iter()
+            .map(|constraint| {
+                constraint.circuit.borrow().evaluate(
+                    main_input.view(),
+                    aux_input.view(),
+                    &challenges,
+                )
+            })
+            .collect_vec();
```
jan-ferdinand, line 1922: I find this easier to follow:

```suggestion
        let output_before_lowering = multicircuit_monad
            .iter()
            .map(|m| m.circuit.borrow())
            .map(|c| c.evaluate(main_input.view(), aux_input.view(), &challenges))
            .collect_vec();
```

Happy about more descriptive variable names than `m` and `c`.

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+            substitution_node_id,
+            0,
+            0,
+        );
+
+        // extract substituted constraint
+        let CircuitExpression::BinOp(BinOp::Add, variable, neg_expression) =
+            &substitution_constraint.circuit.as_ref().borrow().expression
+        else {
+            unreachable!();
```
jan-ferdinand, line 1965: I think a `panic!` is better suited here. While that statement is unreachable as long as the tested method behaves as it should, that's precisely what you're testing here. And while `unreachable!` behaves like `panic!` under the hood, I think the intent here is to communicate that reaching that statement is a bug, and _must_ not happen, as opposed to that it _cannot_ happen.

You could also consider to `use assert2::let_assert` from crate [`assert2`](https://crates.io/crates/assert2):[^dep]

> This [macro] can be used to assert a pattern match, and then run more checks on the captured variables.

```rust
let_assert!(
    CircuitExpression::BinOp(BinOp::Add, variable, neg_expression) =
        &substitution_constraint.circuit.as_ref().borrow().expression
);
```

[^dep]: This crate is already a dev-dependency for other crates in the same workspace.

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+        let extra_input =
+            match &neg_expression.as_ref().borrow().expression {
+                CircuitExpression::BinOp(BinOp::Mul, _neg_one, circuit) => circuit
+                    .borrow()
+                    .evaluate(main_input.view(), aux_input.view(), &challenges),
+                CircuitExpression::BConst(c) => -c.lift(),
+                CircuitExpression::XConst(c) => -*c,
+                _ => {
+                    unreachable!()
+                }
```
jan-ferdinand, line 1976: For a similar reason to above, I think this should be a `panic!`.

```suggestion
                _ => panic!(),
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+        } else {
+            aux_input
+                .append(
+                    Axis(1),
+                    Array2::from_shape_vec([1, 1], vec![extra_input])
+                        .unwrap()
+                        .view(),
+                )
+                .unwrap();
+        }
```
jan-ferdinand, line 1996: I find the following more readable:

```suggestion
        if variable.borrow().evaluates_to_base_element() {
            let extra_input = extra_input.unlift().unwrap();
            let extra_input = Array2::from_shape_vec([1, 1], vec![extra_input]).unwrap();
            main_input.append(Axis(1), extra_input.view()).unwrap();
        } else {
            let extra_input = Array2::from_shape_vec([1, 1], vec![extra_input]).unwrap();
            aux_input.append(Axis(1), extra_input.view()).unwrap();
        }
```

Note that using `extra_input.unlift()` makes use of information hiding: it is an implementation detail how to go from `XFieldElement` to `BFieldElement`.

If you dislike the shadowing, I'm happy about other names. I think this is an acceptable case of variable name shadowing, because (a) the semantics of the variable does not change, and (b) because the shadowed variable gets incorporated into the new binding (what clippy refers to as [`shadow_reuse`](https://rust-lang.github.io/rust-clippy/master/index.html#/shadow_reuse), an allow-by-default lint[^allow]).

[^allow]: This part is a weak argument since `shadow_unrelated`, which I understand you dislike, is also allow-by-default.

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+    /// ```notest
+    ///          C ---- degree-lowering -----> C || k
+    ///          |                               |
+    /// subtract |                      subtract |
+    ///    fixed |                         fixed |
+    ///   output |                        output |
+    ///          |                               |
+    ///          v                               v
+    ///          C* --- degree-lowering ----> C* || k
+    /// ```
```
jan-ferdinand, line 1885: Unfortunately, there are no arrows designed to align nicely with the box drawing characters, as far as I can see.

```suggestion
    /// ```text
    ///          C ──── degree-lowering ─────> C || k
    ///          │                               │
    /// subtract │                      subtract │
    ///    fixed │                         fixed │
    ///   output │                        output │
    ///          │                               │
    ///          v                               v
    ///          C* ─── degree-lowering ────> C* || k
    /// ```
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+    /// and `C` the circuit. Then `input` is a satisfying input for the new circuit
+    /// `C'(X) = C(X) - C(input)`
+    ///
+    /// After applying a substitution to obtain circuit `C || k` from `C`, where
+    /// `k = Z - some_expr(X)` and `Z` is the introduced variable, the length
+    /// of the input and output increases by 1. Moreover, if `input` is a satisfying
+    /// input to `C'` then `input || some_expr(input)` is* a satisfying input to
+    /// `C' || k`.
+    ///
+    /// (*: If the transformation is complete.)
```
jan-ferdinand, line 1861: Are you putting these statements on a separate line to highlight them?

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+    ///
+    /// In this context, completeness means: "given a satisfying assignment to the
+    /// circuit before degree lowering, one can derive a satisfying assignment to
+    /// the circuit after degree lowering." Soundness means the converse.
+    ///
+    /// We test these features using random input vectors. Naturally, the output
+    /// is not the zero vector (with high probability) and so the given input is
+    /// *not* a satisfying assignment (with high probability). However, the
+    /// circuit can be extended by way of thought experiment into one that subtracts
+    /// a fixed constant from the original output. For the right choice of substrahend,
```
jan-ferdinand, line 1848:
```suggestion
    /// a fixed constant from the original output. For the right choice of subtrahend,
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+    /// ```
+    ///
+    /// The point of this elaboration is that in this particular case, testing completeness
+    /// and soundness require the same code path. If `input` and `input || some_expr(input)`
+    /// work for circuits before and after degree lowering, this fact establishes
+    /// both completeness and soundness simultaneously.
+    ///
+    /// Shrinking on this test is disabled because we noticed some weird ass behavior.
+    /// In short, shrinking does not play ball with the arbitrary circuit generator;
+    /// it seems to make the generated circuits *more* complex, not less so.
```
jan-ferdinand, line 1894: In my opinion, because this part of the comment talks about an implementation detail of the test, it should not be a docstring. 

```suggestion
    //
    // Shrinking on this test is disabled because we noticed some weird ass behavior.
    // In short, shrinking does not play ball with the arbitrary circuit generator;
    // it seems to make the generated circuits *more* complex, not less so.
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+                .append(
+                    Axis(1),
+                    Array2::from_shape_vec([1, 1], vec![extra_input])
+                        .unwrap()
+                        .view(),
+                )
+                .unwrap();
+        }
+
+        // evaluate again
```
jan-ferdinand, line 1998: I think this and the next two comments are obvious from the code. If you'd like to have comments here, answering the “why” seems better to me. That said, with the top-level comment in mind, it might not be necessary to repeat.

jan-ferdinand (commented):
The test's top-level comment is very helpful to understand why the test is doing what it is doing, and why it is meaningful to have the test. The test itself is a very good addition.

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1705,316 @@ mod tests {
...
+        let output_after_lowering = multicircuit_monad
+            .iter()
+            .map(|constraint| {
+                constraint.circuit.borrow().evaluate(
+                    main_input.view(),
+                    aux_input.view(),
+                    &challenges,
+                )
+            })
+            .collect_vec();
```
jan-ferdinand, line 2008: Like above.

```suggestion
        let output_after_lowering = multicircuit_monad
            .iter()
            .map(|m| m.circuit.borrow())
            .map(|c| c.evaluate(main_input.view(), aux_input.view(), &challenges))
            .collect_vec();
```

triton-constraint-circuit/src/lib.rs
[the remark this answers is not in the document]
aszepieniec: I modified 66b8225a (now 14a0572) to use newtypes instead of type aliases. Thanks for the reference. Let's proceed with this.

> I see help from the type system to write correct programs as beneficial in all cases.[1](https://github.com/TritonVM/triton-vm/pull/325#user-content-fn-counter-d762d723ea4b4f8dc30c9ddaf200bdf8)

You already read a few general counter-arguments in my bullet point list. In response to this particular phrasing: in the eyes of one person, "help from the type system to write correct programs" is "having to fight the type system to write anything at all in exchange for a marginal assurance of correctness" in the eyes of another. There are programming languages whose type systems are powerful enough to produce a formal computer-verifiable proof of programs' correctness. If rust's type system had this feature, would you recommend using it everywhere and always? The alternative would be to find an optimum on the tradeoff curve interpolating between (a) getting stuff done and (b) correctness.

> >  The function is called in one place only […].
> 
> I don't see how this justifies bad design.

To the extent that the partially elided sentence justifies deviating from your standards (I would obviously not use the B-word) the point is not so much that the function is used in a small number of places and therefore unimpactful; instead, the point made by the elided part is that the function is conceptually and semantically, inseparably part of the of the function that calls it. This relation makes the maintenance of a succinct and misuse-resistant interface moot at best and counter-productive at worst -- because if the algorithm changes then the "interface" will have to adapt.

> Consider:
> 
> ```rust
> let _ = PrintConfig::new(true, true, false);
> let _ = PrintConfig::new(Sided::Single, Colored::Colored, PrintSpeed::Slow);
> ```
> 
> Even though (for arguments sake) all these enums have only two variants, and are thus inherently bools, I strongly prefer the second version.[2](https://github.com/TritonVM/triton-vm/pull/325#user-content-fn-better-d762d723ea4b4f8dc30c9ddaf200bdf8) In my opinion, the argument extends to numerical types.

There is no natural mapping from `Single-Sided`/`Double-Sided` or `Colored`/`Monochrome` or `Slow`/`Fast` to `true`/`false`. You can define a mapping and call it canonical but it would still be an arbitrary choice that preserves no structure. So I disagree with the position that these enums "are inherently bools" and would even argue that the second line is preferable to the first.

To bolster my position that the counts in question *are* inherently `usize`s, consider the necessary boilerplate following the definition of `EvolvingMainConstraintNumber`:
```rust
struct EvolvingMainConstraintsNumber(usize);
impl From<EvolvingMainConstraintsNumber> for usize {
    fn from(value: EvolvingMainConstraintsNumber) -> Self {
        value.0
    }
}
```

In fact, the counts are not used *except* through this `from` method. The type is not used in any algebraic sense.

Maybe that's the point and the essence of our difference. I want types to semantically reflect the underlying maths; whereas you want types to constrain control flow and interfaces.

> In my_vec.remove(i), there's no room for screwing up the argument order.[3](https://github.com/TritonVM/triton-vm/pull/325#user-content-fn-tech-d762d723ea4b4f8dc30c9ddaf200bdf8) The same holds for my_vec.swap(i, j), and so having two arguments of the same type does not influence the method's usability.

There is no room for screwing up the argument order (footnote notwithstanding) but there is room for making the wrong selection from the variables of type `usize` that live in scope. In this case, binding the interface through custom defined types gives the false impression of correctness, which is worse than no impression at all.

triton-constraint-circuit/src/lib.rs
aszepieniec: Further improved with `find_map`.

aszepieniec:
All comments are resolved now (I think).

triton-constraint-circuit/src/lib.rs
```diff
@@ -1471,4 +1716,288 @@ mod tests {
...
+            .cycle()
+            .skip(substitution_node_index % num_nodes)
+            .take(num_nodes)
+            .find_map(|(id, monad)| {
+                if monad.circuit.borrow().is_zero() {
+                    None
+                } else {
+                    Some(id)
+                }
+            })
```
jan-ferdinand, line 1946: Maybe even:

```suggestion
            .find_map(|(&id, monad)| monad.circuit.borrow().is_zero().not().then_some(id))
```

I admit it reads a bit ungrammatical.

jan-ferdinand:
> All comments are resolved now (I think).

I'm still wondering about the following points:

- Can the code in the various `num_` functions be de-duplicated? ([review comment](https://github.com/TritonVM/triton-vm/pull/325/files#r1768739019))
- Is there a copy-paste error in the docstring for `num_nodes`? ([review comment](https://github.com/TritonVM/triton-vm/pull/325/files#r1768714425))
- Is the method how the various `num_` things are computed relevant? ([review comment](https://github.com/TritonVM/triton-vm/pull/325/files#r1768715643))
- Can the hidden relationship between node types in the multicircuit strategy be removed by changing which node types are supplied? ([review comment](https://github.com/TritonVM/triton-vm/pull/325/files#r1768780753))

triton-constraint-circuit/src/lib.rs
aszepieniec: Much more betterer!

jan-ferdinand: I'm honestly not sure it is. :shrug:

triton-constraint-circuit/src/lib.rs
```diff
@@ -1162,6 +1211,133 @@ mod tests {
...
+            constraints
+                .first()
+                .unwrap()
+                .builder
+                .all_nodes
+                .borrow()
+                .len()
+        }
+
+        /// Counts the number of inputs from the main table
```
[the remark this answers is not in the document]
aszepieniec: Fixed in 99424f999303159408e19fd1f2f4db136b0d9476.

triton-constraint-circuit/src/lib.rs
aszepieniec: Yes, but I think the line needs to be changed into `num_additional_nodes`. Even simpler.

aszepieniec:
I think I've answered all your questions. Any further remarks / requests?

jan-ferdinand:
No further requests. Happy to have this test in. :+1:

PR #326, opened 2024-09-18
Thv/faster compilation

Sword-Smith:
Reduces compilation time by 80-90 % by:
- executing build-script code with `opt-level=3`
- caching node-degrees in `ConstraintCircuitMonad::pick_node_to_substitute`

A benchmark of the degree-lowering code on the `thv/degree-lowering-hardening` branch gave us the false impression that degree-lowering was only about 20 seconds. But that was when the code ran with `opt-level=3` which is standard for benchmarks. The default for build scripts is `opt-level=0`. Changing the `opt-level` to 3 speeds up the build script by a factor six. And a further optimization in the degree-lowering code where the node-degrees are cached in the function that is the bottle-neck, as opposed to recalculated many times, gives another 30 % reduction in build times.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,18 @@ lto = "thin"
 [profile.bench]
 lto = "thin"
 
+[profile.release.build-override]
+opt-level = 3
+
+[profile.dev.build-override]
+opt-level = 3
+
+[profile.test.build-override]
+opt-level = 3
+
+[profile.bench.build-override]
+opt-level = 3
+
 [workspace.package]
 version = "0.42.0"
 edition = "2021"
--- a/triton-constraint-circuit/src/lib.rs
+++ b/triton-constraint-circuit/src/lib.rs
@@ -781,25 +781,23 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
         multicircuit: &[ConstraintCircuitMonad<II>],
         target_degree: isize,
     ) -> usize {
-        // The relevant fields of a ConstraintCircuit.
-        // Avoids interior mutability in a HashSet, which is a foot gun.
-        #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
-        struct PartialNodeInfo {
-            id: usize,
-            degree: isize,
-        }
-
         assert!(!multicircuit.is_empty());
         let multicircuit = multicircuit
             .iter()
             .map(|c| c.clone().consume())
             .collect_vec();
 
+        // Computing all node degree is slow; this cache de-duplicates work.
+        let node_degrees = Self::all_nodes_in_multicircuit(&multicircuit)
+            .into_iter()
+            .map(|node| (node.id, node.degree()))
+            .collect::<HashMap<_, _>>();
+
         // Only nodes with degree > target_degree need changing.
         let high_degree_nodes = Self::all_nodes_in_multicircuit(&multicircuit)
             .into_iter()
+            .filter(|node| node_degrees[&node.id] > target_degree)
             .unique()
-            .filter(|node| node.degree() > target_degree)
             .collect_vec();
 
         // Collect all candidates for substitution, i.e., descendents of high_degree_nodes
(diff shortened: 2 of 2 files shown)


Sword-Smith:
Build times on Megingjord (my new desktop):
`master`: 3 minutes 4 seconds
this branch: 19.3 seconds.

Reduction in compilation time: 89.5 %.

triton-constraint-circuit/src/lib.rs
```diff
@@ -781,58 +781,56 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
...
-
         assert!(!multicircuit.is_empty());
         let multicircuit = multicircuit
             .iter()
             .map(|c| c.clone().consume())
             .collect_vec();
 
+        // Precalculate all node degrees, to avoid having to recalculate them
+        // multiple times, which would be a slow affair.
+        let node_degrees = Self::all_degrees(&multicircuit);
```
jan-ferdinand, line 792: I suggest inlining the used-once method `Self::all_degrees()` and shortening the comment.

```suggestion
        // Computing a node's degree is slow; this cache de-duplicates work.
        let node_degrees = Self::all_nodes_in_multicircuit(&multicircuit)
            .into_iter()
            .map(|node| (node.id, node.degree()))
            .collect::<HashMap<_, _>>();
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -781,58 +781,56 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
...
+            .max()
+            .unwrap();
+        candidate_node_ids.retain(|node_id| node_degrees[node_id] == max_degree);
 
         // If there are still multiple nodes, pick any one – but deterministically so.
-        candidate_nodes.sort_unstable_by_key(|node| node.id);
-        candidate_nodes[0].id
+        candidate_node_ids.sort_unstable();
+
+        candidate_node_ids[0]
```
jan-ferdinand, line 833: I'm realizing now that sorting is a potential waste of work, since finding the smallest element has complexity in $\mathsf{O}(n)$.

```suggestion
        candidate_node_ids.into_iter().min().unwrap()
```

triton-constraint-circuit/src/lib.rs
```diff
@@ -872,6 +870,14 @@ impl<II: InputIndicator> ConstraintCircuitMonad<II> {
...
     }
 
+    /// Returns a mapping from node ID to the node's degree.
+    fn all_degrees(multicircuit: &[ConstraintCircuit<II>]) -> HashMap<usize, isize> {
+        Self::all_nodes_in_multicircuit(multicircuit)
+            .into_iter()
+            .map(|node| (node.id, node.degree()))
+            .collect()
+    }
+
```
jan-ferdinand, line 880: Inlining this used-once method keeps things clearer, in my opinion: not two different names for the same thing (`node_degrees` vs `all_degrees`), more type inference, fewer comments needed.

```suggestion
```

Cargo.toml
```diff
@@ -17,6 +17,12 @@ lto = "thin"
 [profile.bench]
 lto = "thin"
 
+[profile.release.build-override]
+opt-level = 3
+
+[profile.dev.build-override]
+opt-level = 3
+
```
jan-ferdinand, line 25: Why not for all the 4 default profiles? Quoting from [the Cargo book](https://doc.rust-lang.org/cargo/reference/profiles.html) (emphasis mine):

> To compile quickly, **all** profiles, by default, do not optimize build dependencies […]

```suggestion
profile.dev.build-override.opt-level = 3
profile.test.build-override.opt-level = 3
profile.bench.build-override.opt-level = 3
profile.release.build-override.opt-level = 3
```

jan-ferdinand (approved):
Good stuff! I left a few {sugg, qu}estions inline.

triton-constraint-circuit/src/lib.rs
Sword-Smith: nice

Cargo.toml
Sword-Smith: cargo gives me warning when I use that syntax. So I opted for:

```toml
[profile.release.build-override]
opt-level = 3

[profile.dev.build-override]
opt-level = 3

[profile.test.build-override]
opt-level = 3

[profile.bench.build-override]
opt-level = 3
```

PR #327, opened 2024-09-18
Thv/bfe xfe arithmetic in tasm code

Sword-Smith:
Exploits both xfe/bfe arithmetic with instructions `xb_mul` and `addi` in the `TasmBackend` code generator. Saves 10 % of the processor table rows in the case of static memory layout and 7.5 % in the dynamic case.

Also no longer pushes an extension field element to the stack when a b-field element suffices.

### DIFF
--- a/triton-constraint-builder/src/codegen.rs
+++ b/triton-constraint-builder/src/codegen.rs
@@ -1,7 +1,9 @@
 //! The various tables' constraints are very inefficient to evaluate if they live in RAM.
 //! Instead, the build script turns them into rust code, which is then optimized by rustc.
 
+use std::cell::RefCell;
 use std::collections::HashSet;
+use std::rc::Rc;
 
 use constraint_circuit::BinOp;
 use constraint_circuit::CircuitExpression;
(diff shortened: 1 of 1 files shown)


triton-constraint-builder/src/codegen.rs
```diff
@@ -743,13 +745,36 @@ impl TasmBackend {
...
+                    BinOp::Add => instr!(XxAdd),
+                    BinOp::Mul => instr!(XxMul),
+                };
+                [lhs, rhs, binop].concat()
+            }
+            (BinOp::Add, None, Some(bfe)) => [lhs, instr!(AddI(bfe))].concat(),
+            (BinOp::Add, Some(bfe), None) => [rhs, instr!(AddI(bfe))].concat(),
+            (BinOp::Mul, None, Some(_)) => [lhs, rhs, instr!(XbMul)].concat(),
+            (BinOp::Mul, Some(_), None) => [rhs, lhs, instr!(XbMul)].concat(),
+        }
```
jan-ferdinand, line 777: I was confused by both comment and code. I suggest the following:[^diff]

```suggestion
        let tokenized_lhs = self.evaluate_single_node(&lhs.borrow());
        let tokenized_rhs = self.evaluate_single_node(&rhs.borrow());
        let tokenized_binop = match binop {
            BinOp::Add => instr!(XxAdd),
            BinOp::Mul => instr!(XxMul),
        };

        // Use more efficient instructions if one side is a base field element.
        // Applying domain-specific knowledge, `CircuitExpression::Input`s can never be
        // base field elements: the constraints are evaluated on out-of-domain rows.
        let extract_bfe_const = |circuit| match circuit.borrow().expression {
            CircuitExpression::BConst(bfe) => Some(bfe),
            _ => None,
        };

        match (binop, extract_bfe_const(lhs), extract_bfe_const(rhs)) {
            (_, Some(_), Some(_)) => panic!("Constant folding should have handled this"),
            (_, None, None) => [tokenized_lhs, tokenized_rhs, tokenized_binop].concat(),
            (BinOp::Add, None, Some(c)) => [tokenized_lhs, instr!(AddI(c))].concat(),
            (BinOp::Add, Some(c), None) => [tokenized_rhs, instr!(AddI(c))].concat(),
            (BinOp::Mul, None, Some(_)) => [tokenized_lhs, tokenized_rhs, instr!(XbMul)].concat(),
            (BinOp::Mul, Some(_), None) => [tokenized_rhs, tokenized_lhs, instr!(XbMul)].concat(),
        }
```

[^diff]: Since the diff includes deleted lines, the interface does not allow me to make a suggestion for the range of lines I suggest to replace. The range starts where the current comment starts and ends as indicated by github's interface.

jan-ferdinand (commented):
Cool optimization. :ok_hand: I was confused by the long `match` expression with nested `match` expression and suggest a simplification of that block inline.

PR #330, opened 2024-09-29
Drop Cached Polynomials

aszepieniec:
This PR drops the cached polynomials. They represent the same information as the original randomized trace, except in a different basis. If they are needed, the relevant randomized trace column is interpolated on the fly.

# Results

I ran benchmark `prove_fib` to gauge performance difference between `master` and this branch. The results are a little confusing.

| fib param | master | this branch |
|---|---|---|
| 25000 | 20.5 Gb / 478 s | 18 Gb / 603 s |
| 50000 | 37.5 Gb / 1013 s | 34.4 Gb / 1281 s |
| 100000 | out-of-memory | out-of-memory |

# Interpretation

The cached polynomials seem to account for only a small fraction of the memory cost, whereas I expected it to be roughly half. The best explanation I can come up with is that the memory-efficient code path is never entered, in which case the low-degree extended trace is cached and this data is roughly 4x larger than the polynomials. But if the memory-efficient code path is never entered, then the slowdown is difficult to explain. (Profiles coming soon.)

If the memory-efficient code path is never entered in the first two rows, then why does the third crash? I would expect that if the low-degree extended trace is not cached, one saves 66% of the memory. So concretely, for growing padded table height I would expect the memory cost to grow until it selects the memory-efficient code path, at which point it drops before growing again, until it crashes. But that does not seem to be happening.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,7 +84,7 @@ strum = { version = "0.26", features = ["derive"] }
 syn = "2.0"
 test-strategy = "0.4.0"
 thiserror = "1.0"
-twenty-first = "0.42.0-alpha.9"
+twenty-first = "0.42.0-alpha.10"
 unicode-width = "0.1"
 divan = "0.1.14"
 
--- a/triton-vm/benches/barycentric_eval.rs
+++ b/triton-vm/benches/barycentric_eval.rs
@@ -6,7 +6,8 @@ use itertools::Itertools;
 use rand::prelude::StdRng;
 use rand::Rng;
 use rand_core::SeedableRng;
-use triton_vm::fri::barycentric_evaluate;
+use twenty_first::math::polynomial::barycentric_evaluate;
+use twenty_first::prelude::XFieldElement;
 
 criterion_main!(benches);
 criterion_group!(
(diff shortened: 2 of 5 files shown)


aszepieniec:
Comparing profiles for Fib 100 one finds almost identical timings except for the out-of-domain rows. On master:
```
├─out-of-domain rows                            2.88s
```
on this branch:
```
├─out-of-domain rows                           36.50s
```

The discrepancy comes from having to use barycentric evaluation instead of Horner evaluation (since we do not have the polynomials anymore). If barycentric evaluation cannot be made any faster, perhaps we should consider cashing the polynomials after all for memory-inefficient code path.

Sword-Smith:
IIRC our current (host machine) Horner evaluation uses memory allocation. In the PR that introduces (host machine) Horner evaluation, I proposed an alternative version that doesn't allocate. It's worth trying that one to see if it solves the problem.

Sword-Smith:
> IIRC our current (host machine) barycentric evaluation uses memory allocation. In the PR that introduces (host machine) barycentric evaluation, I proposed an alternative version that doesn't allocate. It's worth trying that one to see if it solves the problem.

Try implementing barycentric evaluation like [this](https://github.com/TritonVM/triton-vm/pull/266#issuecomment-2072131685) instead. It avoids the allocation and might give a meaningful speedup.

Sword-Smith:
> (...) Some benchmarks, obtain on my laptop. (...)

On mjolnir, only measuring the "out-of-domain rows" step:

| fib parameter | version | time |
|---|---|---|---|
|40 000 | horner | 430.98ms  |
| 40 000 | batch barycentric | 580.05ms|

So even though we see a small slowdown, I think it's worth it to save the RAM and drop the cached interpolants.

Sword-Smith:
**Proving with full caching**

This PR:
```
Prove Fibonacci 40000   time:   [29.757 s 29.974 s 30.182 s]     
```

`master`
```
Prove Fibonacci 40000   time:   [29.104 s 29.216 s 29.317 s]                  
                                           change: [-3.3020% -2.5289% -1.7164%] (p = 0.00 < 0.05)
```

So this PR introduces a 2.5 % slowdown on the fully cached code path.

PR #336, opened 2024-10-09
test: Verify randomizers' large Hamming distance

aszepieniec:
Adds a test that verifies for a dummy aux trace, that all trace randomizers have a sufficiently large Hamming distance.

If this were not so, it would indicate that the RNG is not cryptographically secure or misused.

This test addresses paranoia on my part about rust's `StdRng` whose [documentation](https://docs.rs/rand/latest/rand/rngs/struct.StdRng.html) for `from_seed` points to the [documentation](https://docs.rs/rand_core/latest/rand_core/trait.SeedableRng.html#tymethod.from_seed) for like-named method declared by trait `SeedableRng`, which says this:

> PRNG implementations are allowed to assume that bits in the seed are well distributed. That means usually that the number of one and zero bits are roughly equal, and values like 0, 1 and (size - 1) are unlikely. Note that many non-cryptographic PRNGs will show poor quality output if this is not adhered to. If you wish to seed from simple numbers, use seed_from_u64 instead.

From this I gather that the `SeedableRng` trait does not guarantee that seeding with correlated seeds generates uncorrelated RNGs, but does not forbid implementors from breaking correlation either.

I tried to dig a little into the call graph but before long I stumbled on macro implementations which prevents my IDE from supplying the relevant next links. I did learn that `StdRng` is a newtype wrapper around `BlockRng<ChaCha12Core>`. So `StdRng` is certainly a cryptographically secure random number generator, but it is unclear to me whether there is a cryptographic operation taking place between seeding and generating the first bits.

Turns out, the test passes so there probably is some cryptography happening there. And as best as I can tell, the trace randomizer polynomials are uncorrelated.

That said, the [documentation](https://docs.rs/rand/latest/rand/rngs/struct.StdRng.html) for `StdRng` does say that `ChaCha12` is the *current* implementation and this may change in the future. So if rust ever switches to a new implementation for which correlated seeds generate correlated outputs, this test will catch that and alert the maintainer.

### DIFF
--- a/triton-vm/src/table/master_table.rs
+++ b/triton-vm/src/table/master_table.rs
@@ -2180,14 +2180,13 @@ mod tests {
         print_columns!(aux DegreeLoweringAuxColumn   for "degree low.");
     }
 
-    #[test]
-    fn master_aux_table_mut() {
+    fn dummy_master_aux_table() -> MasterAuxTable {
         let trace_domain = ArithmeticDomain::of_length(1 << 8).unwrap();
         let randomized_trace_domain = ArithmeticDomain::of_length(1 << 9).unwrap();
         let quotient_domain = ArithmeticDomain::of_length(1 << 10).unwrap();
         let fri_domain = ArithmeticDomain::of_length(1 << 11).unwrap();
 
-        let mut master_table = MasterAuxTable {
+        MasterAuxTable {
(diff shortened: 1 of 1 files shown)


jan-ferdinand:
Thanks :bow:

PR #337, opened 2024-10-30
refactor: Introduce structs Prover & Verifier

jan-ferdinand:
Allow producing deterministic proofs by supplying a seed from which all prover randomness is derived.

I'm asking for a review because dealing with randomness incorrectly can break Zero-Knowledge. We currently have no automated mechanism in place to detect broken Zero-Knowledge. (I don't know what such a mechanism would look like.)

This PR has two commits. I recommend reviewing them separately:
- 2f5b7a97805575640046dc9f6324dbf933c854d1 is of lesser importance – feel free to skip it entirely. In this commit, I pave the way by shuffling code in `stark.rs`. Concretely, previously the struct `Stark` contained all helper functions for both, proving and verifying. I introduce two new structs, `Prover` and `Verifier`, and move all proving-related functions into `impl Prover`, and similar for verification-related functions. The methods `Stark::prove` and `Stark::verify` remain in place. They are now pass-through methods, which is a red-flag and indicates a major design issue.[^pass-through] I suggest to refactor the interface to Triton VM in a separate PR in order to keep the scope of this PR succinct.
- adfc20a7b125453de2e082ebd3f9bc6d72103409 is the meat of the PR. I derandomize all proving-related functions, and use a single seed from which all randomness is derived. The `Prover` struct now doesn't implement `Clone` anymore, making accidental re-use of a randomness seed difficult.

[^pass-through]: “A Philosophy of Software Design” by John K. Ousterhout, 2nd Edition, Section 7.1


triton-vm/src/stark.rs
```diff
@@ -947,425 +1381,115 @@ impl Stark {
...
-    /// ⎝  …            …   …             ⎠  ┴
+    /// This method should be the first option to consider for proving.
+    /// For more control over the proving process, see [`Prover`].
+    // This pass-through method guarantees fresh prover randomness at each call.
+    pub fn prove(
+        &self,
+        claim: &Claim,
+        aet: &AlgebraicExecutionTrace,
+    ) -> Result<Proof, ProvingError> {
+        Prover::new(*self).prove(claim, aet)
```
aszepieniec, line 1423: As far as I can tell, this is the only line where something could have gone wrong, and I don't think anything is going wrong.

triton-vm/src/table/master_table.rs
```diff
@@ -575,6 +565,28 @@ where
...
+
+    // entire offset must be used
+    debug_assert!(offset_le_bytes.as_ref().len() <= seed.len());
+
+    // Ensure that the operation is independent of the target pointer:
+    // `to_le_bytes` yields any leading zeros _after_ bits of lesser significance.
+    // Note that this does not guarantee portability across architectures, as
+    // `rand::StdRng` is specifically documented as being not portable.
+    for (seed_byte, offset_byte) in seed.iter_mut().zip(offset_le_bytes) {
+        *seed_byte = seed_byte.wrapping_add(offset_byte);
```
aszepieniec, line 584: Consider xor (marginally faster).

aszepieniec (approved):
Looks good to me.

As best as I can tell, a failure to achieve zero-knowledge would come from seed/rng mis-use (as opposed to the maths, which I think is sound), and I can't find any such mis-use.

Personally, I find the name "`set_randomness_seed_which_may_break_zero_knowledge`" needlessly long and scary. The docstring does an excellent job explaining the caveat emptor. I would suggest using `with_seed` instead.

triton-vm/src/table/master_table.rs
jan-ferdinand: It isn't:

```
Gnuplot not found, using plotters backend
wrapping_add            time:   [372.71 ps 373.97 ps 375.33 ps]
                        change: [-4.9687% -2.9453% -1.2770%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild
  6 (6.00%) high severe

xor                     time:   [377.36 ps 379.01 ps 380.96 ps]
                        change: [-5.2410% -4.5505% -3.8176%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  6 (6.00%) high severe
```

PR #338, opened 2024-11-02
Implement Range for OpStack

### LINKED ISSUE #322: Implement `Index<Range<…>>` for `OpStack`
The operational stack can be indexed into to fetch individual elements through its implementations of `Index<usize>` and `Index<OpStackElement>` (and the corresponding `IndexMut<…>`). It can be useful to get a slice from the op stack.

Implement the combinations {`Index`, `IndexMut`} $\times$ {`Range`, `RangeInclusive`} $\times$ {`usize`, `OpStackElement`} for `OpStack`.

cyberbono3:
Closes #322


jan-ferdinand:
Thanks for taking this on! Unfortunately, there is a logic error in the implementation. The documentation of field `stack` on `OpStack` reads:

```rust
/// The underlying, actual stack. When manually accessing, be aware of reversed indexing:
/// while `op_stack[0]` is the top of the stack, `op_stack.stack[0]` is the lowest element in
/// the stack.
```

Consequently, accessing `op_stack[0..2]` (equivalent to `op_stack.index(0..2)` or `op_stack.index_mut(0..2)` depending on context and determined by the compiler) should give `op_stack.stack[op_stack.stack.len() - 2..op_stack.stack.len()].reverse()` or similar. Note that this probably (a) doesn't compile and (b) is likely to contain off-by-one errors.

triton-air/proptest-regressions/table/processor.txt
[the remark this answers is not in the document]
cyberbono3: I have added `processor.txt` into .gitignore. Does it sound good to you ?

triton-isa/src/op_stack.rs
```diff
@@ -4,6 +4,7 @@ use std::fmt::Result as FmtResult;
 use std::num::TryFromIntError;
 use std::ops::Index;
 use std::ops::IndexMut;
+use std::ops::{Range, RangeInclusive};
```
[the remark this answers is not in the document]
cyberbono3: I have merged all `std::ops` imports together: `use std::ops::{Index, IndexMut, Range, RangeInclusive};`

cyberbono3:
@jan-ferdinand  Sorry for delay. Thanks for your clarifications. Any thoughts on my update?

triton-isa/src/op_stack.rs
```diff
@@ -183,13 +182,69 @@ impl Index<usize> for OpStack {
...
+}
+
+impl Index<Range<usize>> for OpStack {
+    type Output = [BFieldElement];
+
+    fn index(&self, range: Range<usize>) -> &Self::Output {
+        let (start, end) = compute_range(range.start, range.end, self.stack.len());
+        let mut stack = self.stack[start..end].to_vec();
+        stack.reverse();
+        Box::leak(Box::new(stack))
```
jan-ferdinand, line 200: That's a creative way of solving this. :smiley: However, I feel uncomfortable with consciously introducing memory leaks.

The approach you take here indicates to me that implementing `Index<Range<_>> for OpStack` requires refactoring `OpStack` such that the internal field `stack` is always in the correct order, for example using a `VecDeque`. It's probably a good idea to make the field non-public at the same time. Please let me know if this is too much scope creep. :relieved:

jan-ferdinand (changes requested):

triton-isa/src/op_stack.rs
cyberbono3: Introduction of` VecDeque` sounds a good idea.

```
pub struct OpStack {
    pub stack: VecDeque<BFieldElement>,

    underflow_io_sequence: Vec<UnderflowIO>,
}

impl OpStack {
    pub fn new(program_digest: Digest) -> Self {
        let mut stack = VecDeque::with_capacity(OpStackElement::COUNT);
        stack.extend(program_digest.values());

        Self {
            stack,
            underflow_io_sequence: vec![],
        }
    }
```
In this case we do not reverse `program_digest` and set `self.stack[0]` is a top of the stack.

Additionally, we can update `stack.push, stack.pop `to add new item to the beginning of the stack and remove item from the beginning of the stack, respectively. `Vec<BFieldElement>` lacks this functionality.

```
  pub fn push(&mut self, element: BFieldElement) {
        self.stack.push_front(element);
        self.record_underflow_io(UnderflowIO::Write);
    }

    pub fn pop(&mut self) -> Result<BFieldElement> {
        self.record_underflow_io(UnderflowIO::Read);
        self.stack.pop_front().ok_or(OpStackError::TooShallow)
    }

    pub fn insert(&mut self, index: OpStackElement, element: BFieldElement) {
        let insertion_index = usize::from(index);
        self.stack.insert(insertion_index, element);
        self.record_underflow_io(UnderflowIO::Write);
    }
```
`Index<Range<usize>> for OpStack` can be implemented like that :
```
impl Index<Range<usize>> for OpStack {
    type Output = [BFieldElement];

    fn index(&self, range: Range<usize>) -> &Self::Output {
        &self.stack.as_slices().0[range.start..range.end]
    }
}
```

I wonder if this make sense.

triton-isa/src/parser.rs
```diff
@@ -1,12 +1,5 @@
 #![allow(clippy::needless_lifetimes)]
 #![allow(elided_named_lifetimes)]
```
cyberbono3, line 2: I added these attributes to suppress clippy warnings to fix this issues
tritonvm-lifetimeissues

triton-isa/src/op_stack.rs
```diff
@@ -39,9 +39,8 @@ pub const NUM_OP_STACK_REGISTERS: usize = OpStackElement::COUNT;
 #[allow(clippy::len_without_is_empty)]
 pub struct OpStack {
     ///`op_stack[0]` is the top of the stack,
-    /// `op_stack.stack[0]` is the lowest element in the stack. to figure out how it is achieved?
-    // TODO rename this field to queue
-    pub stack: VecDeque<BFieldElement>,
+    /// `op_stack.stack[0]` is the lowest element in the stack.
+    stack: VecDeque<BFieldElement>,
```
cyberbono3, line 43: @jan-ferdinand  would you like to change it to `queue` ?

triton-isa/src/op_stack.rs
```diff
@@ -178,15 +182,44 @@ impl Index<usize> for OpStack {
...
-        &self.stack[top_of_stack - index]
+        &self.stack[index]
+    }
+}
+
+impl Index<Range<usize>> for OpStack {
+    type Output = [BFieldElement];
+
+    fn index(&self, range: Range<usize>) -> &Self::Output {
+        &self.stack.as_slices().0[range.start..range.end]
```
jan-ferdinand, line 193:
```suggestion
        let (front, back) = self.stack.as_slices();

        if front.is_empty() {
            &back[range]
        } else if back.is_empty() {
            &front[range]
        } else {
            panic!("the underlying op stack is not contiguous")
        }
```

jan-ferdinand (changes requested):
It seems we have another fundamental issue going on. :tired_face: Consider the following implementation:

```rust
impl Index<Range<usize>> for OpStack {
    type Output = [BFieldElement];

    fn index(&self, range: Range<usize>) -> &Self::Output {
        let (front, back) = self.stack.as_slices();

        if front.is_empty() {
            &back[range]
        } else if back.is_empty() {
            &front[range]
        } else {
            panic!("the underlying op stack is not contiguous")
        }
    }
}
```

It is super easy to trigger the panic:

```rust
#[test]
fn indexing_with_range_produces_expected_slices() {
    let mut op_stack = test_stack();
    assert_eq!(op_stack[1..3], bfe_array![1, 2]); // 💥
}

/// An [`OpStack`] with [`BFieldElement`] `i` at index `i` and length
/// 2·[`OpStackElement::COUNT`].
fn test_stack() -> OpStack {
    let mut op_stack = OpStack::default();
    for i in (0..OpStackElement::COUNT).rev() {
        op_stack.push(bfe!(i));
    }

    op_stack
}
```

For `impl IndexMut<_> for OpStack`, we can easily avoid this by calling `self.stack.make_contiguous()` in the `index` method. However, for immutable access, this is impossible.

These are the options I see:
- Give up. (😢)
- Only support `IndexMut<_>` in combination with ranges. (😕)
- Use a `VecDeque` and ensure that it is always contiguous. This might impact performance heavily and requires benchmarks.
- Find (or build) a sort of reverse `Vec`.

I'm sorry that this seemingly simple issue is turning into a bit of a rabbit hole. I appreciate the effort you're putting in. At the same time, I want to emphasize that you should feel free to drop this increasingly complex endeavor at any point. :slightly_smiling_face:

cyberbono3:
> It seems we have another fundamental issue going on. 😫 Consider the following implementation:
> 
> ```rust
> impl Index<Range<usize>> for OpStack {
>     type Output = [BFieldElement];
> 
>     fn index(&self, range: Range<usize>) -> &Self::Output {
>         let (front, back) = self.stack.as_slices();
> 
>         if front.is_empty() {
>             &back[range]
>         } else if back.is_empty() {
>             &front[range]
>         } else {
>             panic!("the underlying op stack is not contiguous")
>         }
>     }
> }
> ```
> 
> It is super easy to trigger the panic:
> 
> ```rust
> #[test]
> fn indexing_with_range_produces_expected_slices() {
>     let mut op_stack = test_stack();
>     assert_eq!(op_stack[1..3], bfe_array![1, 2]); // 💥
> }
> 
> /// An [`OpStack`] with [`BFieldElement`] `i` at index `i` and length
> /// 2·[`OpStackElement::COUNT`].
> fn test_stack() -> OpStack {
>     let mut op_stack = OpStack::default();
>     for i in (0..OpStackElement::COUNT).rev() {
>         op_stack.push(bfe!(i));
>     }
> 
>     op_stack
> }
> ```
> 
> For `impl IndexMut<_> for OpStack`, we can easily avoid this by calling `self.stack.make_contiguous()` in the `index` method. However, for immutable access, this is impossible.
> 
> These are the options I see:
> 
> * Give up. (😢)
> * Only support `IndexMut<_>` in combination with ranges. (😕)
> * Use a `VecDeque` and ensure that it is always contiguous. This might impact performance heavily and requires benchmarks.
> * Find (or build) a sort of reverse `Vec`.
> 
> I'm sorry that this seemingly simple issue is turning into a bit of a rabbit hole. I appreciate the effort you're putting in. At the same time, I want to emphasize that you should feel free to drop this increasingly complex endeavor at any point. 🙂

Yeah, you are right. I have identified, that  `IndexMut<_>` with `self.stack.make_contiguous()`  requires  `Index<_> ` implementation. That is impossible to make it work. So I am willing to drop this endeavour.

Regarding reverse` Vec`, from my perspective,  it introduces some additional complexity that vector has to be sorted every time. Thus, the best option is to leave as it is. 

Thanks for your guidance!

jan-ferdinand:
Thank you for trying, and sorry for so many dead ends. :confused:

PR #344, opened 2025-01-08
refactor(verifier): Remove parallelism

jan-ferdinand:
A downstream consumer of Triton VM uses the verifier in an asynchronous context while running its own `rayon` threadpool. This, in combination with Triton VM's use of `rayon`, is suspected to cause odd behavior.

Remove parallelism through explicit use of `rayon` from the Triton VM verifier, with the option to re-introduce it when the underlying issues are better understood.

### DIFF
--- a/triton-vm/src/fri.rs
+++ b/triton-vm/src/fri.rs
@@ -166,8 +166,8 @@ impl ProverRound {
     }
 
     fn merkle_tree_from_codeword(codeword: &[XFieldElement]) -> ProverResult<MerkleTree> {
-        let digests = codeword_as_digests(codeword);
-        MerkleTree::new::<CpuParallel>(&digests).map_err(FriProvingError::MerkleTreeError)
+        let digests: Vec<_> = codeword.par_iter().map(|&xfe| xfe.into()).collect();
+        MerkleTree::par_new(&digests).map_err(FriProvingError::MerkleTreeError)
     }
 
     fn split_and_fold(&self, folding_challenge: XFieldElement) -> Vec<XFieldElement> {
@@ -385,7 +385,6 @@ impl FriVerifier<'_> {
         let folding_challenge = round.folding_challenge.unwrap();
 
         (0..self.num_collinearity_checks)
-            .into_par_iter()
             .map(|i| {
                 let point_a_x = domain.domain_value(a_indices[i] as u32).lift();
                 let point_b_x = domain.domain_value(b_indices[i] as u32).lift();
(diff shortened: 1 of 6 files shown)


Sword-Smith:
This PR's branch, without parallelism:
```
Verify Halt             time:   [8.1461 ms 8.1681 ms 8.1947 ms]
```

`master`
```
time:   [11.644 ms 11.680 ms 11.735 ms] 
```

Removing parallelism makes verification faster, because of the overhead of rayon thread-handling.

triton-vm/src/fri.rs
```diff
@@ -166,7 +166,7 @@ impl ProverRound {
     }
 
     fn merkle_tree_from_codeword(codeword: &[XFieldElement]) -> ProverResult<MerkleTree> {
-        let digests = codeword_as_digests(codeword);
+        let digests: Vec<_> = codeword.par_iter().map(|&xfe| xfe.into()).collect();
         MerkleTree::new::<CpuParallel>(&digests).map_err(FriProvingError::MerkleTreeError)
```
aszepieniec, line 170: We will need a single-threaded Merkle tree builder too, no?

jan-ferdinand: The line you highlight is in the prover. I have removed the parallelism from the function `codeword_as_digests` and inlined the parallel version that is used in the prover.

jan-ferdinand: Oh, I think I misunderstood your comment. Let me check the code again.

jan-ferdinand: You are absolutely right. The FRI verifier is building a Merkle tree from the last round's codeword, and the only logic for building Merkle trees we currently have is using parallelism.[^1]

[^1]: Getting into the nitty gritty details, upon which the FRI verifier should absolutely not depend, Merkle trees with fewer than 512 leafs are built in sequence (unless a rather specific environment variable overrides the default).

As a consequence, we seem to require a new Merkle tree builder, `CpuSequential`. Would you agree?

aszepieniec: > we seem to require a new Merkle tree builder, CpuSequential. Would you agree?

Yes please!

I also noticed one place in some guesser where we use that. Since it's only 8 leafs there, it probably defaults to sequential anyway. It would be good to call the right method directly though, and not have to rely on contextual knowledge.

jan-ferdinand: It seems that the code for FRI is not as deduplicated as it could be; `merkle_tree_from_codeword` is a function associated with `ProverRound`, and is never called in the verifier, even though the verifier performs the same operations in its method `last_round_codeword_merkle_root`. Thanks for the heads up, though!

jan-ferdinand:
Blocked: waiting for twenty-first v0.45.0.

Issue #278, opened 2024-05-07
Parallelize Constraints Generation


aszepieniec:
.

aszepieniec:
Done in 0ac5c370590f5dbfa58c58ddc3aa471c7bf6d208.

PR #357, opened 2025-06-11
Can a builder be just a `Vec`?

skaunov:
After tinkering enough with the crates and <build.rs> I'd like to discuss this tiny example of a simpler approach to making the builder. It has pasted and modified code of `triton-constraint-circuit` to reflect the change approach and just compile, and a single adapted test to start with and have a discussion on moving forward.

When describing this task @aszepieniec had a suggestion of adding like `Mutex` for parallelization, though to me it feels better to play this simpler if possible and just go with less fancy `Vec`. This degrades readability and elegance of the code (and I don't see how to really fix this), but improves parallelization while leveraging the compiler safety checks, and more naturally reusing constants (like $-1$ or other).

Sadly both approaches requires tedious rewrite of a lot of code in most of the crates here. Tbh I would start either of this with adding as much tests as possible. I mean coverage is cool already, but rewriting/refactoring of this kind seems to be always error-prone so additional tests would only help.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "triton-vm",
 ]
 resolver = "2"
+exclude = ["simpler"]
 
 [profile.test]
 opt-level = 1
--- a/simpler/Cargo.toml
+++ b/simpler/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "plain"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+triton-constraint-circuit = {path = "../triton-constraint-circuit"}
+itertools = "0.14"
+ndarray = { version = "0.16", features = ["rayon"] }
+rand = "0.9.0"
+proptest = "1.6"
+test-strategy = "0.4.0"
+proptest-arbitrary-interop = "0.1"
+arbitrary = { version = "1", features = ["derive"] }
+twenty-first = "0.48.0"
+num-traits = "0.2"
+quote = "1.0"
+proc-macro2 = "1.0"
--- a/simpler/src/lib.rs
+++ b/simpler/src/lib.rs
@@ -1,0 +1,1058 @@
+//! Constraint circuits are a way to represent constraint polynomials in a way
+//! that is amenable to optimizations. The constraint circuit is a directed
+//! acyclic graph (DAG) of [`CircuitExpression`]s, where each
+//! `CircuitExpression` is a node in the graph. The edges of the graph are
+//! labeled with [`BinOp`]s. The leafs of the graph are the inputs to the
+//! constraint polynomial, and the (multiple) roots of the graph are the outputs
+//! of all the constraint polynomials, with each root corresponding to a
+//! different constraint polynomial. Because the graph has multiple roots, it is
+//! called a “multitree.”
+
+use std::cmp;
+
+use std::fmt::Debug;
+use std::fmt::Display;
+use std::fmt::Formatter;
+use std::fmt::Result as FmtResult;
+use std::hash::Hash;
+use std::hash::Hasher;
+
+use std::ops::Add;
+
+use std::ops::Mul;
+
+
+
+
+use arbitrary::Arbitrary;
+use ndarray::ArrayView2;
+use num_traits::One;
+use num_traits::Zero;
+use quote::ToTokens;
+use quote::quote;
+use twenty_first::prelude::*;
+
+mod private {
+    // A public but un-nameable type for sealing traits.
+    pub trait Seal {}
+}
+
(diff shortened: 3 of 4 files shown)


jan-ferdinand:
I think this approach has merit. It also has downsides.

#### Pros

- **Simplicity**. In particular, I like that there are no (transitively) recursive data structures anymore.
- **Speed Potential**. As far as I can tell, it should be possible to speed up the bottleneck of compilation (being `pick_node_to_substitute`) with this approach. It might be worth to translate that function as one of the next steps to see if the approach translates nicely.

#### Cons

- **Exposition of Internals**. I don't like how the constraint builder volunteers its internal indices, and how it accepts plain old indices to perform its operations. I'd prefer to have a new type, akin to `CircuitNode`.[^name] This type could be a wrapper around the plain index, keeping it private. It would also mean that it's impossible to accidentally pass something unrelated as an index to the circuit builder. If, additionally, it's impossible to construct the type from outside the `constraint-circuit` crate, that would mean that whatever is being passed to a builder is some circuit node. Things might still be able to go out of sync, but it'd be more difficult.
- **Ergonomics**. Having to write `builder.add(left, right)` instead of `left + right` is a bummer. It might be possible to bring ergonomics back by teaching the suggested `CircuitNode` about the builder that built it. However, this might well be in contradiction to the **Simplicity** I see as a pro. After all, if a builder knows its nodes _and_ the nodes know their builder, that sounds like a circular dependency, which your current approach does not have. It might be that we have to choose between either ergonomics when using the builder and simplicity when writing the builder.

#### Additional Nice-to-Haves

- Both in the current and in your approach, roots of the multicircuit are just a loose collection of nodes. Sometimes, they are collected in an array or vector. It might be beneficial to have a dedicated type, maybe `MultiCircuit`, that ties everything together. This might help to prevent things to go out of sync. I say this now mostly so it's out there; I don't think it has the highest priority at the moment, and it might not even belong with your PR at all.

[^name]: this name is a placeholder

skaunov:
Thanks for the hint for the next spot to try out!

Passing `usize` around is bad, I was thinking mainly about an alias; but if you'd elaborate this _CircuitNode_ idea it would be great.

Yeah, using vs writing is quite a core of this. One of the angles I wanted to take is start with a converter which would just take a current Builder before generating the code and convert it to a simpler one to do parallelization. But the code generation itself is really integrated with that circular structure. As now I know builders surely can't mix I might look into this another.

Btw, would you think it would be a decent speed up if the code for the three files in <build.rs> will be generated in parallel. Or most of the problem is before that point? 

Also want to link this to #278 .

jan-ferdinand:
> elaborate this _CircuitNode_ idea

I was thinking of something like the following.

```rust
#[derive(Debug, Clone, Eq, PartialEq)]
pub struct ConstraintCircuitBuilder<II: InputIndicator> {
    nodes: Vec<CircuitNode<II>>,
}

/// A node in the multicircuit. Should only be used together with the builder
/// from which it originated.
//
// In order to guarantee that only actual nodes are given to the
// `ConstraintCircuitBuilder` when building up a multicircuit, only the builder
// can create `CircuitNode`s. This does not prevent things from going out of
// sync completely, since two different builders could be involved. If a future
// redesign wants to protect against that, builders could have a unique ID
// that they pass on to the nodes they create, and which could be checked
// when the node is used in a context that requires a builder.
#[derive(Debug, Copy, Clone, Eq, PartialEq)]
pub struct CircuitNode<II: InputIndicator> {
    /// The index at which to find this node in its builder's
    /// [`nodes`][ConstraintCircuitBuilder::nodes].
    index: usize,

    /// The circuit expression represented by this node. In a sense, this is
    /// the actual node.
    circuit: ConstraintCircuit<II>,
}

// It might make sense to also implement `Deref` for `CircuitNode`, but that
// should be re-evaluated. It's quite possible that there's a better design.
impl<II: InputIndicator> Deref for CircuitNode<II> {
    type Target = ConstraintCircuit<II>;

    fn deref(&self) -> &Self::Target {
        &self.circuit
    }
}
```

The `ConstraintCircuitBuilder` can then be refactored to never expect any indices directly, and to never hand them out, either.

> One of the angles I wanted to take is start with a converter which would just take a current Builder before generating the code and convert it to a simpler one

My gut feeling says that this would have been a painful path to walk down. :sweat_smile: It's probably _possible_, but I don't think it's easier than re-writing everything.

> would you think it would be a decent speed up if the code for the three files in <build.rs> will be generated in parallel. Or most of the problem is before that point?

I think that's a good idea and worth trying out. I know that degree lowering is slow, but there might be other slow parts.

skaunov:
I only see `pick_node_to_substitute_is_deterministic` test on it. Is there other helpful things possible to do with such a limited translation?
> * It might be worth to translate that function as one of the next steps to see if the approach translates nicely.

PR #358, opened 2025-06-16
chore: Update license to MIT OR Apache-2.0

jan-ferdinand:


LICENSE-APACHE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2021 Triton Software AG
```
aszepieniec, line 189: Shouldn't that be 2025?

aszepieniec (approved):

LICENSE-APACHE
jan-ferdinand: Maybe. I decided to go with the repository's initialization year, deciding against the other obvious options of “2025” and “2021-2025” because (a) [it seems to not matter](https://opensource.stackexchange.com/questions/5778/) and (b) if the license lists no current year, it doesn't go out of date as quickly.
That said, I don't have a strong opinion in the matter. 🤷

aszepieniec: The source you link to suggests that, to the extent that it does matter (which is mostly "not really"), the first relevant date is preferable. And the argument about 2021 going out of date less quickly is well taken. Let's keep it as is.

PR #373, opened 2025-09-08
feat: implement program hash integrity constraints for continuations

avorylli:
Add mathematical constraints to prevent program hash overwriting via swap instructions when opstack size ≤ 21.

- Prevents swap operations from targeting program hash elements (st11-st15) in shallow stacks
- Provides formal constraint: `swap_indicator · (n - 21) · indicator_{i≥11} = 0`
- Resolves TODO in TIP-0008 section (a) Program

### DIFF
--- a/tips/tip-0008/tip-0008.md
+++ b/tips/tip-0008/tip-0008.md
@@ -55,22 +55,9 @@ The only way in which the program hash could be overwritten is through `swap`, `
 
 **Constraints.** To prevent the program hash from being overwritten by stack manipulation instructions when the opstack size is 21 or less, we need to ensure that these operations cannot target the program hash elements (st11 through st15) when the stack is shallow enough to expose them.
 
-Let $n$ be the current opstack size and let $i$ be the target index. The constraint is:
+The constraint implementation depends on the specific arithmetic circuit framework used by Triton VM. For practical implementation, we suggest using a range check approach where the stack size $n$ is constrained to be within valid bounds, and stack manipulation operations targeting indices $i \geq 11$ are only allowed when $n > 21$.
 
-$$\text{stack\_manip\_indicator} \cdot \text{indicator}_{n \leq 21} \cdot \text{indicator}_{i \geq 11} = 0$$
-
-where:
-- $\text{stack\_manip\_indicator}$ is 1 when executing a `swap`, `pick`, or `place` instruction, 0 otherwise
-- $\text{indicator}_{n \leq 21}$ is 1 when $n \leq 21$ (stack is shallow enough to expose program hash), 0 otherwise
-- $\text{indicator}_{i \geq 11}$ is 1 when $i \geq 11$ (targeting program hash elements), 0 otherwise
-
-This constraint ensures that when the stack size is 21 or less ($n \leq 21$), any stack manipulation operation targeting program hash elements ($i \geq 11$) will be invalidated by the constraint evaluation.
-
-The indicator function $\text{indicator}_{n \leq 21}$ can be implemented as an arithmetic circuit of low degree using the fact that $n$ is bounded by the maximum stack size. Specifically, we can use the constraint:
-
-$$\text{indicator}_{n \leq 21} = \prod_{k=22}^{n_{max}} (n - k)$$
-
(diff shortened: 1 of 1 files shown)


aszepieniec (commented):
Thanks for the contribution.

The instructions `pick` and `place`, which were introduced after this note was written, likewise have the capacity to disturb the program hash. So they should be included also in `swap_indicator`.

>  $(n - 21)$ is positive when $n > 21$ (stack is deep enough to protect program hash)

I'm afraid I don't see how the factor $n-21$ activates the constraint when $n \leq 21$. Rather, it looks to me like it *dis*activates the constraint when $n = 21$ and activates it in all other cases.

I do see how the factor `max(0, 21-n)` achieves this purpose, but then the question is how to represent `max` as an arithmetic circuit of low degree.

tips/tip-0008/tip-0008.md
```diff
@@ -51,9 +51,22 @@ The state of the VM is fully determined by a bunch of things, divisible into the
...
 
-The only way in which the program hash could be overwritten is through `swap` instructions when the opstack size is 21 or less. 
+The only way in which the program hash could be overwritten is through `swap`, `pick`, and `place` instructions when the opstack size is 21 or less. 
 
-**Constraints.** Todo.
+**Constraints.** To prevent the program hash from being overwritten by stack manipulation instructions when the opstack size is 21 or less, we need to ensure that these operations cannot target the program hash elements (st11 through st15) when the stack is shallow enough to expose them.
+
+Let $n$ be the current opstack size and let $i$ be the target index. The constraint is:
+
+$$\text{stack\_manip\_indicator} \cdot (21 - n) \cdot \text{indicator}_{i \geq 11} = 0$$
```
aszepieniec, line 60: I'm afraid this constraint is not complete. Counterexample: the instruction is `swap 15` when $n = 22$. The constraint evaluates to $P-1 \neq 0$ meaning that this particular instruction in this particular context is not allowed. But it *should be* allowed because when $n = 22$ there is no risk of disturbing the program hash.

aszepieniec (changes requested):

tips/tip-0008/tip-0008.md
```diff
@@ -51,9 +51,26 @@ The state of the VM is fully determined by a bunch of things, divisible into the
...
+where:
+- $\text{stack\_manip\_indicator}$ is 1 when executing a `swap`, `pick`, or `place` instruction, 0 otherwise
+- $\text{indicator}_{n \leq 21}$ is 1 when $n \leq 21$ (stack is shallow enough to expose program hash), 0 otherwise
+- $\text{indicator}_{i \geq 11}$ is 1 when $i \geq 11$ (targeting program hash elements), 0 otherwise
+
+This constraint ensures that when the stack size is 21 or less ($n \leq 21$), any stack manipulation operation targeting program hash elements ($i \geq 11$) will be invalidated by the constraint evaluation.
+
+The indicator function $\text{indicator}_{n \leq 21}$ can be implemented as an arithmetic circuit of low degree using the fact that $n$ is bounded by the maximum stack size. Specifically, we can use the constraint:
+
+$$\text{indicator}_{n \leq 21} = \prod_{k=22}^{n_{max}} (n - k)$$
```
aszepieniec, line 71: I think it is correct now, but you've introduced another drawback: the maximum stack size is actually $P-1 = 2^{64} - 2^{32}$ which is *huge* when it is supposed to be a bound on the degree of an AIR constraint.

aszepieniec (changes requested):

avorylli:
made it simpler, if you don't like it, then probably I could think of something else to improve the previous variant

jan-ferdinand:
Frankly, I don't think protecting the program's hash digest against modification is necessary. In my mind, the entire program is executed by one entity. There is never any “execution handover”; for example, it is not possible to execute a different program as a subroutine of some sort (other than straight up incorporating that other program, then executing it, which means that it's not a different program).

Summarizing, with the threat model we're currently working with, I don't see additional protection we gain by disallowing write access to the program's hash digest.

In case I misunderstand your, @avorylli, intentions, would you mind opening an issue? There might be other possible solutions to discuss before deciding on a strategy to implement them.

PR #385, opened 2025-10-18
docs: Add deriving-air-constraints.md with a concise, reusable derivation process.

### LINKED ISSUE #273: Docs: Describe Derivation of Constraints; Don't Present Them
Currently, the AIR constraints documentation is difficult to read and even more difficult to maintain. The constraints, as presented, are not informative. We should not be presenting the them but rather describing where they came from and how to (re-)derive them.

Note that someone who is unfamiliar with our codebase might be a little confused about how a description maps onto a constraint. So we need an additional section explaining how to derive a concrete set of constraints from a textual description of what they are supposed to do.

MozirDmitriy:
fix #273 

- Added specification/src/deriving-air-constraints.md with a concise, reusable derivation process.
- Linked it in specification/src/SUMMARY.md.
- Updated specification/src/lookup-table.md to add a Derivation section and moved equations under “Appendix: Polynomial Forms”.
- Updated specification/src/jump-stack-table.md similarly.

### DIFF
--- a/specification/src/jump-stack-table.md
+++ b/specification/src/jump-stack-table.md
@@ -137,6 +137,16 @@ Let all household items (🪥, 🛁, etc.) be challenges, concretely evaluation 
 Let all fruit & vegetables (🥝, 🥥, etc.) be challenges, concretely weights to compress rows, supplied by the verifier.
 Both types of challenges are X-field elements, _i.e._, elements of $\mathbb{F}_{p^3}$.
 
+## Derivation
+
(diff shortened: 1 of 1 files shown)


jan-ferdinand:
This is not quite what the issue you point to means. Is the new content LLM-generated? I highly doubt this approach will work, because (a) all LLM's I've interacted with so far have been terrible at math and (b) I've never seen any written description of the derivation process the issue is alluding to, meaning that the required training material for an LLM probably doesn't exist.

PR #390, opened 2025-11-06
docs: Decrease width of spec page “instructions”

jan-ferdinand:
Also:
- add table summarizing each instruction
- briefly list the (informal) instruction categories
- break out section “about instructions”
- introduce README.md to `./specification`
- put instruction `push` before instruction `pop`
- correctly describe padding behavior of instruction `hash`
- describe general `ip` increment behavior
- remove some superfluous horizontal lines before footnotes
- add more inter-doc links

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,20 @@ generate or verify proofs of correct execution, or
 check out the [Triton CLI](https://github.com/TritonVM/triton-cli). If you want to use Triton VM as
 a library, check out the [examples](triton-vm/examples).
 
+## Project Status
+
+We consider Triton VM to be usable in production.
+That said, there are plans to change some of the internals significantly.
+Some of these changes might have effects on the interface as well as the ISA.
+
+As is common for rust projects, Triton VM follows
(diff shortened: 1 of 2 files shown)


specification/src/instructions.md
```diff
@@ -1,182 +1,811 @@
...
+[arithmetization](instruction-specific-transition-constraints.md#instruction-skiz)).
+The second property helps with proving consistency of the
+[Op Stack](data-structures.md#operational-stack).
+The third property allows efficient arithmetization of the running product for the
+Permutation Argument between [Processor Table](processor-table.md) and [U32 Table](u32-table.md).
+
+## Summary
+
+The following table is a summary of all instructions.
+For more details, read on below.
```
skaunov: Summary is really helpful! Would you also mention before the table all the sectioning used below? (As we don't know if we will ever have a ToC here.) I feel the minimal comma separated naming of each section in a line would be enough, though there are nicer ways too. X)

jan-ferdinand: Good idea. Done. :+1:

PR #397, opened 2025-12-04
Try to improve compile times: reduce allocations when building

jan-ferdinand:
Benchmarks indicate that a large chunk of the time required to build Triton VM is due to vector allocations in `ConstraintCircuitMonad::all_nodes_in_circuit`. By implementing a custom iterator over `ConstraintCircuit`s, I ~~hope~~ manage to reduce the number of required allocations and thus to improve build times.


jan-ferdinand:
> Merging #397 will **improve performances by ×5**

Promising, promising. The fact that the number of nodes as reported in the arithmetization overview has changed indicates that something funky might be going on.

PR #405, opened 2026-01-09
Use STIR as the low-degree test

jan-ferdinand:
Add STIR as an alternative to FRI.

When reviewing, please pay special attention to parameter derivation. There are a bunch of things here that I don't know how to test for.
The file `fri.rs` has moved and undergone _some_ changes, but not nearly as much as the diffs suggests. In particular, the prover is completely unchanged, and the only change to the verifier is that it also constructs a `VerifierPostcript`. When reviewing, feel free to check that this postcript contains all information necessary for the Triton assembly implementation. The rest of `fri.rs` can be skimmed or even ignored, I think.

### DIFF
--- a/triton-vm/benches/proof_size.rs
+++ b/triton-vm/benches/proof_size.rs
@@ -1,7 +1,5 @@
 use std::collections::HashMap;
-use std::ops::AddAssign;
 
-use criterion::BenchmarkGroup;
 use criterion::BenchmarkId;
 use criterion::Criterion;
 use criterion::Throughput;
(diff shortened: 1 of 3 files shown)


triton-vm/src/profiler.rs
```diff
@@ -611,9 +613,12 @@ impl VMPerformanceProfile {
         self
     }
 
+    /// The length of the low-degree test's initial domain.
+    ///
+    /// Adds additional information when set.
     #[must_use]
-    pub fn with_fri_domain_len(mut self, fri_domain_len: usize) -> Self {
-        self.fri_domain_len = Some(fri_domain_len);
+    pub fn with_low_deg_test_domain_len(mut self, ldt_domain_len: usize) -> Self {
```
aszepieniec: "LDT" is a perfectly acceptable commonplace abbreviation for "low degree test", and indeed one that is used elsewhere within this codebase. Suggested refactor: `with_ldt_domain_len`.

triton-vm/benches/proof_size.rs
```diff
@@ -181,11 +181,12 @@ fn program_halt() -> ProgramAndInput {
     ProgramAndInput::new(triton_program!(halt))
 }
 
-/// The base 2, integer logarithm of the FRI domain length.
-fn log_2_fri_domain_length(stark: Stark, proof: &Proof) -> u32 {
+/// The base 2, integer logarithm of the length of the low-degree test domain.
+fn log_2_low_deg_test_domain_length(stark: Stark, proof: &Proof) -> u32 {
```
aszepieniec: "LDT" is a perfectly acceptable commonplace abbreviation for "low degree test", and indeed one that is used elsewhere within this codebase. Suggested refactor: `log_2_ldt_domain_length`.

triton-vm/src/stark.rs
```diff
@@ -1540,24 +1524,37 @@ impl Stark {
...
+        // Also, the degree of the trace-randomized interpolation polynomial for
+        // any trace column must be less than the length of the randomized trace
+        // domain.
+        // Together, these conditions imply:
+        //
+        //    len > degree(randomized_trace_column_interpolant)
+        //  ⟺ len > num_trace_randomizers + degree(zerofier(trace_domain))
+        //  ⟺ len > num_trace_randomizers + (len / 2)
+        //  ⟺ len > 2·num_trace_randomizers
+        let smallest_fit_for_rand_trace_interpolation_poly = 2 * num_trace_randomizers + 1;
```
aszepieniec: I understand the comment. I do not see how it gives rise to this variable. Maybe the link is not apparent because of the name. I see a derivation of a minimum length. What does that have to do with a polynomial? And why does the polynomial have type `usize`?

Suggestion: `min_length_for_rand_trace_interpolation`.

triton-vm/src/stark.rs
```diff
@@ -1569,27 +1566,84 @@ impl Stark {
         max_degree_supported_by_that_smallest_arithmetic_domain as isize
     }
 
-    /// The parameters for [FRI](Fri). The length of the
-    /// [FRI domain](ArithmeticDomain) has a major influence on
-    /// [proving](Prover::prove) time. It is influenced by the length of the
-    /// [execution trace](AlgebraicExecutionTrace) and the FRI expansion factor,
-    /// a security parameter.
+    /// The low-degree test, [STIR](Stir).
```
aszepieniec: If this function is the low degree test STIR, then what does it return? I would naively assume that the object of type `Stir` is the low degree test STIR, but that obviously can't be true if this function already occupies that description.

triton-vm/src/stir.rs
```diff
@@ -0,0 +1,2060 @@
...
+    /// this STIR instance.
+    ///
+    /// In other words, the low-degreeness of polynomials with degree
+    /// 2^log2_high_degree and higher cannot be [proven](Stir::prove) (in a way
+    /// that the [verifier](Stir::verify) accepts).
+    ///
+    /// Must be greater than or equal to the (log₂ of the)
+    /// [folding factor](Self::log2_folding_factor).
+    #[cfg_attr(test, strategy(#log2_folding_factor..=15))]
+    pub log2_high_degree: usize,
```
aszepieniec: Are you using the word "high" as in the opposite of "low" as in "low degree test"? That's rather non-standard. People coming with background knowledge of low degree tests understand that they distinguish polynomials whose degree is less than some bound from polynomials whose degree is larger. (With an "or equal" inserted somewhere.) In other words, it's "low degree or not low degree" as opposed to "low degree or high degree".

I suggest to change to: `log2_degree_bound`.

triton-vm/src/stark.rs
```diff
@@ -1569,27 +1566,84 @@ impl Stark {
...
+        // doubling in the domain size grants ample additional space if the
+        // `padded_height` is sufficiently large. It's only for very small
+        // `padded_heights` that additional iterations are necessary, and there,
+        // things are fast anyway.
+        let log2_padded_height = usize::try_from(padded_height.ilog2()).expect(U32_TO_USIZE_ERR);
+        let log2_max_domain_growth_factor = ArithmeticDomain::LOG2_MAX_LEN - log2_padded_height;
+
+        for log2_domain_growth_factor in 1..=log2_max_domain_growth_factor {
+            let mut parameters = stir_parameters;
+            parameters.log2_high_degree += log2_domain_growth_factor;
```
aszepieniec: Why add 2, 3, 4, etc. in every iteration and not add 1 in every iteration?

triton-vm/src/stark.rs
```diff
@@ -1569,27 +1566,84 @@ impl Stark {
...
-            self.num_collinearity_checks,
-        )
+        // The initial domain of STIR must be longer than the length of the
+        // randomized trace (after accounting for the expansion factor).
+        // However, the STIR instance itself influences the number of trace
+        // randomizers, and good luck writing out the direct dependency between
+        // the `log2_high_degree` and the number of trace randomizers. I'm sure
+        // it can be done, I'm not sure if it's worth the effort. Instead, we
+        // perform a linear search for the smallest factor by which to grow the
+        // initial domain such that it is big enough.
```
aszepieniec: This non-trivial dependency calls into question whether migrating to STIR is a good idea. What if we need a bigger domain relative to if we had stuck with FRI?

Perhaps, and most hopefully, the domain we need for STIR is always smaller than or at most equal to the domain we need for FRI. If this is true, then we might as well use the FRI domain as a starting point and ask whether we can shrink it.

triton-vm/src/stir.rs
```diff
@@ -0,0 +1,2060 @@
...
+    //
+    // Using this to upper-bound the probability Pr[D < k] still has a somewhat
+    // unwieldy summation going on:
+    //
+    //    Pr[D < k] ⩽ Σ_(d=1)^(k-1) (U choose d) · (d/U)^n
+    //
+    // We upper-bound this further by simply replacing each summand with the
+    // largest of them all:
+    //
+    //    Pr[D < k] ⩽ (k-1) · (U choose k-1) · ((k-1)/U)^n
```
aszepieniec: This does not work for very small U (in particular, when U-k > k) but I guess we are not working in that domain, right?

aszepieniec (changes requested):
I reviewed the parameter selection algorithms carefully (excluding list size) and I think they make sense. Or at least, the ones that relate to soundness make sense. There is a bug, I think, in the number of trace randomizer rows calculation.

I also reviewed the prover carefully and I think it is correct. If the verifier behaves right in relation to that, it must be correct as well.

triton-vm/src/stark.rs
jan-ferdinand: I'm not sure how to meaningfully build on top of a previous iteration's result here. That said, I agree that it's easier to understand what's going on when the field in question is incremented by 1 in every iteration.

triton-vm/src/stir.rs
```diff
@@ -0,0 +1,2060 @@
...
+        Polynomial::new(folded_coefficients)
+    }
+
+    fn next_round_domain(domain: ArithmeticDomain) -> ArithmeticDomain {
+        let next_domain = domain
+            .pow(1 << StirParameters::LOG2_DOMAIN_SHRINKAGE)
+            .unwrap();
+
+        next_domain.with_offset(next_domain.offset() * domain.offset())
+    }
```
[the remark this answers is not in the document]
jan-ferdinand: Because `StirParameters::LOG2_DOMAIN_SHRINKAGE` is a private constant unequal to 0. We can also add a
```rust
const { assert_ne!(0, StirParameters::LOG2_DOMAIN_SHRINKAGE) };
```
somewhere, for example right next to it, but that seems a little redundant at the current stage. Is your point rather that the potential future refactor that generalizes the domain shrinkage should make sure that the domains actually shrink between rounds?

triton-vm/src/low_degree_test/fri.rs
```diff
@@ -6,25 +7,78 @@ use twenty_first::math::traits::FiniteField;
...
+    pub log2_initial_expansion_factor: usize,
+
+    /// The low-degree test's degree bound.
+    ///
+    /// In particular, the (log₂ of the) polynomial degree that is considered
+    /// “high” (_i.e._, “not low”) for this FRI instance.
+    ///
+    /// In other words, the low-degreeness of polynomials with degree
+    /// `2^log2_high_degree_bound` (and higher) cannot be [proven](Fri::prove)
+    /// (in a way that the [verifier](Fri::verify) accepts). On the other hand,
```
aszepieniec:
```suggestion
    /// (in a way that the [verifier](Fri::verify) accepts whp.). On the other hand,
```

triton-vm/src/low_degree_test/stir.rs
```diff
@@ -0,0 +1,2020 @@
+//! The [STIR](Stir) polynomial low-degree test over the
+//! [extension field](XFieldElement).
+//
+// Future enhancements may include:
+//
+// - A folding factor that varies between rounds.
+// - “Domain shrinkage” that varies between rounds.
+// - Giacomo et al.: “Interactive Proofs for Batch Polynomial Evaluation”. This
+//   makes interpolation in the verifier superfluous.
```
aszepieniec: Are you referring to what we call "Giacomo's trick" in the office? Because as I recall, Giacomo's trick allows getting rid of interpolation in the verifier, but I don't see what that has to do with batch polynomial evaluation or how that's an interactive oracle proof. Maybe there is more than one unused trick due to Giacomo in circulation?

triton-vm/src/low_degree_test/stir.rs
```diff
@@ -0,0 +1,2020 @@
...
+            let log2_folded_poly_degree = folded_poly_degree.ilog2();
+            let out_of_domain =
+                self.num_ood_queries(log2_folded_poly_degree, log2_next_expansion_factor)?;
+
+            let num_queries = NumQueries {
+                in_domain,
+                out_of_domain,
+            };
+
+            // In STIR's quotienting step, the so-called “answer” polynomial is
```
aszepieniec: I think it's called $\mathsf{Ans}$ in the paper but agree that "answer" is the most likely disabbreviation.

triton-vm/src/low_degree_test/fri.rs
```diff
@@ -575,7 +784,105 @@ impl Fri {
...
+impl Fri {
+    /// (Try to) construct a new FRI instance for the given initial parameters.
+    ///
+    /// It is equivalent to use the provided trait implementation of
+    /// [`TryFrom<FriParameters> for Fri`](TryFrom).
+    ///
+    /// # Errors
+    ///
+    /// Errors if any of the pre-conditions documented in the [`FriParameters`]
+    /// are violated.
```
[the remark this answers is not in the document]
jan-ferdinand: You're right. FRI is just _that_ much simpler than STIR, it seems.

triton-vm/src/low_degree_test/stir.rs
```diff
@@ -0,0 +1,2020 @@
...
+        // polynomial of degree 0. However, folding a polynomial of degree 3
+        // _also_ results in a polynomial of degree 0, even though the degree-3
+        // polynomial should be rejected by the verifier. The solution is to
+        // stop folding before the process “bottoms out.” Since there is one
+        // more folding step applied in the final round, no full round must
+        // reduce the degree to be less than or equal to the folding factor.
+        //
+        // Also note that a STIR instance with a “high degree” of less than the
+        // folding factor is degenerate and will have been rejected earlier in
+        // this method.
```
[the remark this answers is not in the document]
jan-ferdinand: I also thought about dynamically modifying the folding factor for the last round to be smaller than initially requested to prevent “bottoming out” the polynomial's degree. For example, if the current polynomial's degree is 4, a folding factor of 4 could be used, even though a folding factor of 8 or higher might have been requested.

I suspect that this approach has similar benefits to your suggestion while being more performant (but that's just a guess). The definite drawback is the need to support folding factors that vary between rounds, which we can't do yet.

PR #415, opened 2026-02-04
ci: Work around eprint.iacr.org rate limit

jan-ferdinand:
Concretely, stop checking for broken links to eprint.iacr.org.

Private communications with the site admin indicate that the eprint server is under constant heavy load, largely due to so-called AI agents. I certainly don't want to contribute to their problems in any way. A few broken links here and there are not ideal but better than pummeling their infrastructure even more.


jan-ferdinand:
Apparently, some even-more-recent changes have made the mitigations I attempted here superfluous.

PR #432, opened 2026-05-27
Univariate batching

jan-ferdinand:
Use a single, randomly sampled indeterminate per batching “category” instead of column-many indeterminates to batch the combination codeword.

Also, replace instruction `xb_dot_step` with `b_horner_step` and `xx_dot_step` with `x_horner_step`.


specification/src/instructions.md
```diff
@@ -709,37 +708,54 @@ Crashes the VM if `i` is not u32.
...
 
 **Opcode**: 80
 
-Reads two extension field elements from RAM located at the addresses corresponding to the two top
-stack elements, multiplies the extension field elements, and adds the product `(p0, p1, p2)` to an
-accumulator located on stack immediately below the two pointers.
-Also, increases the pointers by the number of words read.
+This instruction facilitates efficient evaluation of a univariate polynomial over the base field in an indeterminate
+from the extension field.
+In order to leverage this instruction most efficiently, the polynomial should be encoded with it's leading coefficient
```
aszepieniec, line 717:
```suggestion
In order to leverage this instruction most efficiently, the polynomial should be encoded with its leading coefficient
```

specification/src/instruction-specific-transition-constraints.md
```diff
@@ -929,24 +929,24 @@ See also [`merkle_step`](#instruction-merkle_step).
...
 ### Description
 
-1. Store `RAM[st0]` in `hv0`.
-1. Store `(RAM[st1], RAM[st1+1], RAM[st1+2])` in `(hv1, hv2, hv3)`.
-1. Add `hv0 · (hv1 + hv2·x + hv3·x²)` into `(st1, st2, st3)`
-1. Increase the pointers: `st0` and `st1` by 1 and 3, respectively.
+1. Store `RAM[st5]` in `hv2`, `RAM[st5 - 1]` in `hv1`, and `RAM[st5 - 2]` in `hv0`.
+1. Stack elements `st0` through `st4` as well as `st6` do not change.
+1. Decrease RAM pointer `st5` by 3.
+1. Move `(st7 + st8·x + st9·x²) · (st7 + st8·x + st9·x²) + (hv0 + hv1·x + hv2·x²)` into `st7` through `st9`.
```
aszepieniec:
```suggestion
1. Move `(st0 + st1·x + st2·x²) · (st7 + st8·x + st9·x²) + (hv0 + hv1·x + hv2·x²)` into `st7` through `st9`.
```

I'm not sure if `st0` through `st2` is correct but `(st7 + st8·x + st9·x²) · (st7 + st8·x + st9·x²)` seems wrong. Same mistake above in `b_horner_step`.

triton-air/src/table/processor.rs
```diff
@@ -2014,140 +2013,103 @@ fn xb_product<Indicator: InputIndicator>(
...
-    .map(curr_main_row);
-    let hv_product = xx_product([hv0, hv1, hv2], [hv3, hv4, hv5]);
+    // compress current and next stack
+    let stack_weight = |i| circuit_builder.challenge(stack_weight_by_index(i));
+    let curr_stack_compressed = stack_weight(0) * curr_main_row(MainColumn::ST0)
+        + stack_weight(1) * curr_main_row(MainColumn::ST1)
+        + stack_weight(2) * curr_main_row(MainColumn::ST2)
+        + stack_weight(3) * curr_main_row(MainColumn::ST3)
+        + stack_weight(4) * curr_main_row(MainColumn::ST4)
+        + stack_weight(5) * (curr_main_row(MainColumn::ST5) - constant(1))
```
aszepieniec, line 2106: The -1 is because we are incrementing the ram pointer by 1, right? (Perhaps worth a comment.)

triton-vm/src/stark.rs
```diff
@@ -2121,7 +2121,7 @@ impl<'a> Arbitrary<'a> for Stark {
 
 /// Fiat-Shamir-sampled challenges to compress a row into a single
 /// [extension field element][XFieldElement].
```
aszepieniec, line 2141: Worthwhile mentioning: univariate batching means all generated from one element; for performance, this struct caches all powers of that generator.

triton-vm/src/stark.rs
```diff
@@ -2135,24 +2135,35 @@ struct LinearCombinationWeights {
...
+
+        let all_weights =
+            |initial: XFieldElement, num| (0..num).map(|i| initial.mod_pow_u32(i)).collect();
+        let main = all_weights(initial_weight_main_columns, num_main_columns);
+        let aux = all_weights(initial_weight_aux_columns, num_aux_columns);
+        let quot_segments = all_weights(initial_weight_quot_segments, num_quot_segments);
+        let deep = all_weights(
+            initial_weight_deep_codeword_components,
+            num_deep_codeword_components,
+        );
```
aszepieniec: Setting the DEEP component weights to the next powers of the same singular generator strikes me as difficult to prove sound at best, and insecure at worst. The reason is the operands for this second step of batching were already compressed using powers of the generator. Using a power derived from the same generator might interfere and cancel out the first step. I would sample another scalar, and use the powers of this second scalar for batching the DEEP components.

aszepieniec (changes requested):
I found some pickable nits in the docs.

Deriving the 3 DEEP component weights from powers of the same scalar whose powers are used to compress the low-degree extended trace columns is worth flagging. It is not obviously secure to me.

In fact, from the point of view of simplifying the soundness proof the most, it pays to sample 4 different scalars.
 - 1 for the main table
 - 1 for the aux table
 - 1 for the quotient table
 - 1 for the DEEP components.

triton-air/src/table/processor.rs
jan-ferdinand: We're actually decrementing the RAM pointer since the `BFieldCodec`-encoding of `Polynomial<FF>` puts the polynomial's leading coefficient at the largest address.[^order] Because an iterative application of Horner's method requires starting with the leading coefficient, decrementing the RAM pointer seemed like a good idea.

[^order]: pretending, for a moment, that field elements are ordered

specification/src/instruction-specific-transition-constraints.md
jan-ferdinand: Good catch. `st0` through `st2` is correct.

triton-vm/src/stark.rs
```diff
@@ -2129,7 +2139,17 @@ impl<'a> Arbitrary<'a> for Stark {
...
 /// Fiat-Shamir-sampled challenges to compress a row into a single
 /// [extension field element][XFieldElement].
-struct LinearCombinationWeights {
+///
+/// Rows are compressed using “univariate batching”, that is, the entries of a
+/// row are interpreted as the coefficients of a univariate polynomial. The
+/// “batching” step then evaluates that polynomial in a given indeterminate. In
+/// the case of the STARK, the indeterminates are “supplied by the verifier.”
+/// (Given the non-interactive nature of the STARK achieved by the BCS
+/// transform, the indeterminates are sampled using the Fiat-Shamir heuristic.)
```
aszepieniec: It's strange and confusing to mention the BCS transform here. That's the step where you replace codewords with their Merkle roots, and later simulate all oracle-queries with authentication paths. I don't think it has any bearing on how the indeterminates are sampled.

triton-vm/src/stark.rs
```diff
@@ -2143,24 +2163,49 @@ struct LinearCombinationWeights {
...
+        let initial_weight_quot_segments = initial_weights[1];
+        let initial_weight_deep = initial_weights[2];
+
+        // Pre-compute the needed powers of the various initial weights.
+        // Note that across the main & aux table, no two columns must use the
+        // same weight. Consequently, exponent 0 must be used at most once
+        // between the `main` and `aux` category.
+        let powers = |x: XFieldElement, range: Range<_>| range.map(|i| x.mod_pow_u32(i)).collect();
+        let main = powers(initial_weight_main, 0..num_main_columns);
+        let aux = powers(initial_weight_aux, 1..num_aux_columns + 1);
```
aszepieniec:
```suggestion
        let aux = powers(initial_weight_msin, num_main_columns..num_main_columns + num_aux_columns + 1);
```

The weights for the $M+A$ columns should be $(1, X, \ldots, X^{M+N-1})$ not $(1, X, \ldots, X^{M-1}, X^M, X^{2M}, X^{3M}, \ldots, X^{AM})$.

aszepieniec (changes requested):
Now, I do think the construction is obviously sound. However, there is a construction that

 - is simpler
 - is even more obviously sound
 - has a concretely smaller soundness error.

triton-vm/src/stark.rs
jan-ferdinand: Oh yeah, I meant the Fiat-Shamir heuristic.

triton-vm/src/stark.rs
jan-ferdinand: Good catch. :+1:

jan-ferdinand:
> However, there is a construction that [is better]

Is it the one where `let aux = powers(initial_weight_main, num_main_columns..num_main_columns + num_aux_columns);` or a different one still?

aszepieniec (approved):

aszepieniec:
> > However, there is a construction that [is better]
> 
> Is it the one where `let aux = powers(initial_weight_main, num_main_columns..num_main_columns + num_aux_columns);` or a different one still?

No, that's the one.

PR #434, opened 2026-06-11
Enforce that executed opcodes are valid

jan-ferdinand:
Previously, it was possible to execute non-existent instructions, allowing for arbitrary state transitions. I currently rank the criticality of such an event as medium-low since either
1. the verifier learns the program and can check that it only contains valid instructions, or
1. the verifier doesn't learn the program, in which case the program can already behave arbitrarily.

Nonetheless, it doesn't _feel_ right that the verifier has to check opcodes for their validity when they do learn the program. In particular, not performing the required check (for example due to lack of knowledge or an oversight) can lead to surprising and potentially unpleasant consequences.

The commit in this pull request enforces that any executed (!) instruction is valid. It:
1. changes the instruction deselectors such that their image is $\\\{0, 1\\\}$, and
1. introduces one additional constraint, $\sum_\texttt{instr} (\text{delector}_\texttt{instr}) - 1$, enforcing that exactly one deselector is inactive.[^dn]

[^dn]: don't you just not hate the double negatives?


aszepieniec (approved):
LGTM

PR #438, opened 2026-06-15
Fix air soundness gaps

aszepieniec:


jan-ferdinand (approved):
Looks perfect. 👌

triton-air/src/table/accumulator_pinning.rs
jan-ferdinand, line 1: I don't currently agree with the inclusion of `accumulator_pinning.rs`. Several stylistic issues aside (see below), the tests included therein feel like they help the process of identifying gaps in the AIR, but they also seem to be something of a clutch. The knowledge required to set up and test for any given suspected gap in the AIR strikes me as highly specific, preventing this approach from becoming more general. I currently think we're better of to use the results produced by these tests, but to not include the tests and write something more general instead.


### Stylistic issues

Code & comments break with the style established in the rest of the codebase. To name a few:

- Functions like `generic_challenges` are duplicated code. (In this specific case, use `Challenges::default` instead.) Other places construct challenges themselves, calling neither of the now multiple available functions.
- Comments introduce new terminology without explaining it, like “guard column” or  “off-domain value”.
- There is dead code (even marked `#[allow(dead_code)]`).
- `cargo fmt` produces a non-empty diff.
- Several comments describe old behavior or talk about specific changes, for example “<-- the fix”.

triton-air/src/table/hash.rs
```diff
@@ -1056,7 +1068,18 @@ impl AIR for HashTable {
...
+                // `round_number_next` is ≥ 0 over the legal range and zero only
+                // at rn' = 0, so the mode guard must also be ≥ 0 and zero only at
+                // mode' = Hash. Using `mode_next - Hash` (≤ 0) instead gives the
+                // two guards opposite signs, so their sum cancels at the reachable
+                // interior rows where rn' + mode' = 3 — (ProgramHashing, rn' = 2)
+                // and (Sponge, rn' = 1) — leaving the running evaluation free.
+                // Hence `Hash - mode_next`, mirroring the digest constraint's
+                // same-sign guards `(rn' - NUM_ROUNDS) + (mode' - Hash)`.
+                + (circuit_builder.b_constant(HashTableMode::Hash) - mode_next.clone())
                     * running_evaluation_hash_input_remains;
```
jan-ferdinand, line 1083: A comment in the middle of an expression (a) is odd and (b) prevents `cargo fmt` from doing its job.
Given that we're using both `Self::round_number_deselector` and `Self::mode_deselector`, I don't even think a comment regarding the `Self::select_mode` introduced in this PR is necessary.

```suggestion
        let running_evaluation_hash_input_is_updated_correctly =
            Self::round_number_deselector(circuit_builder, &round_number_next, 0)
                * Self::mode_deselector(circuit_builder, &mode_next, HashTableMode::Hash)
                * running_evaluation_hash_input_updates
                + round_number_next.clone() * running_evaluation_hash_input_remains.clone()
                + Self::select_mode(circuit_builder, &mode_next, HashTableMode::Hash)
                    * running_evaluation_hash_input_remains;
```

triton-air/src/table/op_stack.rs
```diff
@@ -85,10 +85,24 @@ impl AIR for OpStackTable {
...
+        let main_row =
+            |column: Self::MainColumn| circuit_builder.input(Main(column.master_main_index()));
+
+        // Confine `IB1ShrinkStack` to its legal set {0, 1, PADDING=2} on every
+        // row. Without this, the column is constrained only indirectly through
+        // the permutation argument, whose coefficient of
+        // `RunningProductPermArg_next` is `ib1_next² − 2`; that coefficient
+        // vanishes at the out-of-set value `ib1_next = √2` (which exists since
+        // the prime is 1 mod 8), freeing the perm-arg terminal at the
+        // real→padding boundary.
```
jan-ferdinand, line 100:
```suggestion
```

triton-air/src/table/ram.rs
```diff
@@ -97,10 +97,25 @@ impl AIR for RamTable {
...
+        let main_row =
+            |column: Self::MainColumn| circuit_builder.input(Main(column.master_main_index()));
+
+        // Confine `InstructionType` to its legal set {WRITE=0, READ=1,
+        // PADDING=2} on every row. Without this, the column is constrained only
+        // indirectly through the permutation argument, whose coefficient of
+        // `RunningProductPermArg_next` is `type_next² − 2`; that coefficient
+        // vanishes at the out-of-set value `type_next = √2` (which exists since
+        // the prime is 1 mod 8), freeing the perm-arg terminal at the
+        // real→padding boundary.
```
jan-ferdinand, line 112:
```suggestion
```

triton-air/src/table/ram.rs
```diff
@@ -97,10 +97,25 @@ impl AIR for RamTable {
...
+        let instruction_type = main_row(Self::MainColumn::InstructionType);
+        let instruction_type_is_legal = instruction_type.clone()
```
jan-ferdinand, line 114:
```suggestion
        let instruction_type_is_legal = (instruction_type.clone()
            - constant(INSTRUCTION_TYPE_WRITE))
```

jan-ferdinand (commented):
While I'm a big fan of regression tests, I'm not sure (a) the many, highly specific, rather verbose tests are what we want in this codebase, and (b) they should be included in the branch called `critical_fixes`.

How about we include the fixes here and do the tests properly on the main branch?

